PR #76, opened 2015-05-05
Fix CI Build

JackMc:
Fix the CI build by re-using a container if we can in the run_container method. This adds a lookup (equivalent to "docker ps") into the run_container method.


wardi:
alternate fix in 19b8c54

PR #80, opened 2015-05-07
background task support for shell and paster commands

deniszgonjanin:


wardi:
nice!

wardi:
Can you enable circleci on your fork so that your PRs get tested?

deniszgonjanin:
Ooh, fancy. I will look into that

PR #82, opened 2015-05-09
Allow multiple "child environments" under a single environment

JackMc:
This PR adds the ability to create multple child environments under an environment you create, sharing the same source and venv but not the same database/web/files.

The changes to the datadir here include each child environment having its own subdirectory for its containers and the passwords.ini containing a secret key for each child env.

Note that this had not been tested very heavily on Linux (this would be appreciated, including any feedback. I intend to do it on Monday). 

Also note that automatic migration of existing projects/datadirs to the new format isn't entirely working on Mac because I don't know how to move a data-only container. Any tips?

TOFIX
- [x] Fix option parameter to look like the rest of the parameters i.e. `-c NAME` instead of `--child=<name>` in the docstring
- [x] Make migration work on Mac (move data-only containers)
- [ ] Test more thoroughly on Linux (this seems to be why at least 'purge' is failing in CircleCI in certain cases)


wardi:
https://docker-py.readthedocs.org/en/latest/api/#rename looks like the thing we need for moving data-only containers

datacats/cli/create.py
```diff
@@ -20,7 +20,7 @@ def create(opts):
     """Create a new environment
 
 Usage:
-  datacats create [-bin] [--ckan=CKAN_VERSION] ENVIRONMENT_DIR [PORT]
+  datacats create [-bin] [-c NAME] [--ckan=CKAN_VERSION] ENVIRONMENT_DIR [PORT]
```
wardi: I'm torn on this.

Should we allow the user to choose a child name on create or just force them to create "primary" and use init to add children. Can you think of a reason to create a new environment and custom child name at the same time?

JackMc:
Thanks, Ian! Looks like the purge failure on Linux was being caused by the fact that postgres was still being stored in the main datadir on Linux because I forgot to move the folder.

Also, is there some way we can test boot2docker in the CircleCI?

datacats/cli/manage.py
```diff
@@ -171,14 +178,15 @@ def open_(environment, opts):
...
+  datacats open [-r] [-c NAME] [ENVIRONMENT]
 
 Options:
   -r --remote        Open DataCats.com cloud instance address
+  -c --child=NAME  Choose a child environment to open [default: primary]
 
 ENVIRONMENT may be an environment name or a path to an environment directory.
 Default: '.'
-"""
+t"""
```
wardi: typo?

datacats/cli/create.py
JackMc: Well, in the use case where you want unique names for each of your sub-environments. That was the use case I was thinking of, but I can see the merit of doing it in "init" also.

datacats/cli/manage.py
JackMc: Yes. Thanks, I'll fix that! :)

datacats/cli/create.py
wardi: Yeah, my concern is a new user will see the docs and be confused about what the child name thing is. I can appreciate the uniformity of having the option on all commands where it makes sense though. As I said, I'm torn :-)

datacats/environment.py
```diff
@@ -1,10 +1,10 @@
-# Copyright 2014-2015 Boxkite Inc.
+
```
wardi: typo

JackMc: Fixed

datacats/environment.py
```diff
@@ -57,31 +57,56 @@ def __init__(self, name, target, datadir, ckan_version=None, port=None,
...
+        self.children = None
 
-    def save(self):
+    def _load_children(self):
         """
-        Save environment settings into environment directory
+        Loads child names for this env from the .datacats-environment file.
+        This sets the property 'children' of the Environment.
+        This method is safe to call as long as a valid .datacats-environment
+        exists (it doesn't need a 'datacats/children' directive).
```
wardi: Let's move the children in `datadir + '/children'` and then get this list from a listdir() call instead. The .datacats-environment file's children won't make sense if this file is shared to another user or machine.

I fully admit there are other settings in .datacats-environment that also might not make sense when shared, but let's not make things worse. :-)

JackMc: I agree

JackMc: Should ports be kept in the .datacats-environment file still, or should we make them generated on calls to 'start' (for example)

datacats/environment.py
```diff
@@ -130,20 +178,106 @@ def new(cls, path, ckan_version, port=None):
...
-    def load(cls, environment_name=None, data_only=False):
+    def _needs_format_conversion(cls, datadir):
+        """
+        Returns True if `datadir` requires conversion to the new format (child envs).
+        """
+        return isdir(datadir) and isdir(join(datadir, 'run')) and exists(join(datadir, 'passwords.ini')) and exists(
+                join(datadir, 'search')) and exists(join(datadir, 'solr')) and exists(join(datadir, 'project-dir'))
+
+    @classmethod
+    def _convert_environment(cls, datadir):
```
wardi: Please start a new migrations.py module for this code. A classmethod that never uses `cls` screams "this should be a normal function"! :-)

JackMc: Sounds good!

datacats/environment.py
```diff
@@ -130,20 +178,106 @@ def new(cls, path, ckan_version, port=None):
...
+        new_child_name = 'primary'
+        inp = None
+
+        while inp != 'y' and inp != 'n':
+            inp = raw_input('You are using a file in the old DataCats format. Would you like to convert it (y/n) [n]: ')
```
wardi: How about a `datacats migrate` command instead. We can have the other commands fail on load() with an error saying "this environment is out of date, please run "datacats migrate" to upgrade it to the latest version."

datacats/environment.py
wardi: It's nice to remember the port the user requested so that we don't move around each time. Refreshing a web browser is much nicer on a "datacats reload". I think we should move them to a setting in the datadir.

wardi:
boot2docker won't run in circleci, but it might run in travis or another CI that gives us root.

JackMc:
I'll look into it

datacats/environment.py
JackMc: I mean shouldn't the port be the same given the same environment name and child name though? It's deterministic IIRC

JackMc:
To make future conversions easier, do you reckon we should version our datadir formats? Currently I just check for a whole bunch of directories.

datacats/environment.py
wardi: The first port chosen is deterministic, but the user is allowed to specify a port they like instead on create and start. Also, if the port is already in use datacats will keep trying new ports until it finds a free one (I'm not sure if the one it lands on is saved, but it probably should be)

wardi:
versioning datadirs is a great idea. We should also have an exclusive lock that we take when performing migrations: take the lock, ensure all containers are stopped, and load() should fail while the lock is present.

datacats/environment.py
JackMc: Instead of storing it we could always determine it from the web container on-reload

JackMc:
CircleCI passes! ^_^

wardi:
moving the project might have broken the circleci status thing for me, I see no status on this PR any more. My google-fu is failing and I don't see anything in the circleci settings :-(

JackMc:
It borked the indicator in the README, but I merged something this morning to fix it (changed boxkite->datacats). I dunno what else needs to be done.

JackMc:
Merged the latest in - may be a little unstable

wardi:
@JackMc Great! Has the rest settled down? should I push out a release and start the new master by merging this one?

JackMc:
Depends if you want the --address thing and the lessc commands in the new release - they need some polish

JackMc:
If you're okay with waiting for the next release for those features, then I'd say that's a good idea!

JackMc:
I've played with this for a little bit and CircleCI'd it and it seems to work with all the new features and not break existing child functionality

JackMc:
More master merges

wardi:
- remove the backup created during migration (we're just moving directories)
- report on the version that we migrated to "migrated to data format 2"
- add a migrate option to downgrade "--format 1" or similar (copy the primary dirs back to where they were) so someone can revert to an earlier datacats for testing or other purposed
- add downgrading and upgrading to the circleci tests, with a check in between that the old directories exist where they used to

wardi:
- please reword 'If you interrupted a migration, your best course of action would be to do  "datacats purge"' to suggest manual recovery (remove the lock file?) or use purge as a last resort removing all data and files

wardi:
I get a failure on datacats init:
```python-traceback
datacats init -c two
Creating environment registry23/two from existing environment directory "registry23"
  File "/home/ian/boxkite/venv/bin/datacats", line 9, in <module>
  File "/home/ian/boxkite/git/datacats/datacats/cli/main.py", line 80, in main
    return command_fn(opts)
  File "/home/ian/boxkite/git/datacats/datacats/cli/create.py", line 148, in init
    fn()
  File "/home/ian/boxkite/git/datacats/datacats/cli/create.py", line 137, in <lambda>
    lambda: environment.create_directories(create_project_dir=False)] + ([
  File "/home/ian/boxkite/git/datacats/datacats/environment.py", line 382, in create_directories
    makedirs(self.datadir + '/venv')
  File "/home/ian/boxkite/venv/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/home/ian/.datacats/registry23/venv'

```

JackMc:
@dborzov @wardi Please CR :)

wardi:
I haven't reviewed this yet but since we're talking about names on another PR, these should probably be "sites" (à la "multi-site") not "children"

JackMc:
Do you mean all over or just in a specific bit?

wardi:
No change yet. Let's talk names tomorrow

JackMc:
Okie dokie

dborzov:
Obviously we are calling them dataKittens

JackMc:
- [ ] More thorough CircleCI tests.

JackMc:
- [ ] Rename "children" to "sites"

JackMc:
This is ready for CR/testing (again) @wardi @dborzov

wardi:
broke it with `datacats migrate -r 0` :smile:

wardi:
`datacats init -s test` causes install to be run even if another site already exists, which is a little wasteful

`datacats info` should split up its information by site, e.g.
```
Environment name: registry23
    Default port: 5888
 Environment dir: /home/ian/environments/registry23
        Data dir: /home/ian/.datacats/registry23

            Site: primary
      Containers: web postgres solr
    Available at: http://127.0.0.1:5888/

            Site: test
      Containers: web postgres solr
    Available at: http://127.0.0.1:5492/
```

wardi:
`datacats deploy` doesn't need a `--site` option. It deploys from the source directory.

`datacats logs` is failing with a KeyError: '--solr-logs' (Need to add a test to circle CI)

`datacats open` is using `-c` for the short `--site` option and doesn't allow selecting a non-default site

`datacats paster` and `datacats shell` do not work with `-s`

`datacats purge -s primary` when there are other sites followed by `datacats info` on linux shows ` Data dir: /home/ian/.datacats/registry23 (damaged)`

JackMc:
Thanks for testing all this :smile:

JackMc:
The issue with ``datacats shell`` and ``datacats paster`` seems to be a deeper problem with the way we parse command line arguments related to those commands. They do not account for the fact that a command could be separated by a space from its argument and thus insert the -- something like this:

``['shell', 'testenv', '-s', '--', 'devel']``

When they should be something like this:
``['shell', 'testenv', '-s', 'devel', '--']``

There seems to be no way for us to actually tell the difference between an argument to a - option and a positional argument save actually parsing the docstring. This means our method of inserting a -- may now not be sufficient.

Suggestions?

JackMc:
The only option I can think of that works without more docopt hacks is making it so that users must specify their shell parameters in a single positional argument (i.e. they must quote their commands if they're more than one word).

The other option is to take a sh-like approach to it and have a "-c" option on the shell command. I think this is more explicit and consistent with the way people expect things like this to work. Also we can remove the hacky stuffs :stuck_out_tongue:

JackMc:
That's everything on the lists you gave me @wardi, thanks for the input :smile: If you find anything else, let me know!

wardi:
purge seems to be leaving behind a venv directory

PR #84, opened 2015-05-10
A Couple Style Fixes

JackMc:
This makes the code compatible with pep8, assuming we're conforming to that standard.

Please feel free to close if this is not necessary, there are a couple tabbing fixes in there which Ian told me to fix.


datacats/cli/main.py
```diff
@@ -81,7 +83,7 @@ def main():
         if help_:
             args = ['--help']
         return docopt(__doc__, args,
-            version=pkg_resources.require("datacats")[0].version)
+                      version=pkg_resources.require("datacats")[0].version)
```
wardi: this is my one quibble with pep8: don't make me align to opening parens. this creates all sorts of noise when the first line changes and the rest need to be re-indented.

datacats/environment.py
```diff
@@ -255,9 +261,9 @@ def data_complete(self):
         """
         Return True if all the expected datadir files are present
         """
-        if (not isdir(self.datadir + '/files')
-                or not isdir(self.datadir + '/run')
-                or not isdir(self.datadir + '/search')):
+        if (not isdir(self.datadir + '/files') or
+                not isdir(self.datadir + '/run') or
+                not isdir(self.datadir + '/search')):
```
wardi: oh, I also don't like this. the "or" in this continuation is very important for understanding the logic so I prefer it at the beginning of the line near the rest of the logical operators.

wardi:
Please add pep8 checking to the tests too, with exceptions for the bits I commented on, and I'm happy to merge.

Have you tried flake8 or pylint? Might be something to consider too for later.

JackMc:
Okie dokie will do that in the morning!

wardi:
No rush on any of this of course! And linting is super-low priority. I think these changes boil down to adding blank lines after skipping down the stuff I've commented on.

JackMc:
Closing to open an updated version.

PR #85, opened 2015-05-11
Fix CircleCI indicator so it points at new repo

JackMc:
(It still pointed at boxkite/datacats, which doesn't exist anymore)


JackMc:
Spoke with Denis

Issue #87, opened 2015-05-12
Allow listening on more than 127.0.0.1 on linux


wardi:
Add an option to start/reload that binds to a custom address (including 0.0.0.0) when running CKAN.

wardi:
A workaround with the current version is to edit https://github.com/datacats/datacats/blob/master/datacats/environment.py#L585 and change '127.0.0.1' to '0.0.0.0'

PR #88, opened 2015-05-13
Add --address option to reload and start to listen on a given host

JackMc:
Linux-only, can specify 0.0.0.0 to listen on all interfaces.

Addresses #87


datacats/environment.py
```diff
@@ -568,6 +569,7 @@ def _run_web_container(self, port, command):
             ro = {self.datadir + '/venv': '/usr/lib/ckan'}
             volumes_from = None
 
+        print 'Binding on host {}'.format(host)
```
wardi: Don't print from environment.py, do that in the cli module instead. And here only do it if the user specifies a non-default value

wardi:
How about `--address` instead? Host is often used for a host name, and wouldn't that always be the same?

wardi:
Also, please make it an error to specify a host if is_boot2docker() is True (raise an EnvironmentError)

datacats/environment.py
JackMc: That was a debug line, sorry.

JackMc:
Changed title to reflect new option name

datacats/cli/manage.py
```diff
@@ -33,13 +33,14 @@ def start(environment, opts):
...
-  datacats start [-bp] [ENVIRONMENT [PORT]]
-  datacats start -r [-b] [ENVIRONMENT]
+  datacats start [-bp] [-a HOST] [ENVIRONMENT [PORT]]
+  datacats start -r [-b] [-a HOST] [ENVIRONMENT]
 
 Options:
   -b --background    Don't wait for response from web server
   -p --production    Start with apache and debug=false
   -r --remote        Start DataCats.com cloud instance
+  -a --address=HOST   Host to listen on (Linux-only) [default: 127.0.0.1]
```
wardi: `--address=IP` in the example is a little clearer I think

JackMc: Ok

JackMc:
I'm also worried that users may be confused about the short form (-a IP), because it's similar to the typical "all" short form.

wardi:
good thought. And we might want -a for something later. Listening on a different interface is obscure enough to not need a short form, just make users say `--address`

JackMc:
Sounds good!

datacats/environment.py
```diff
@@ -489,11 +489,12 @@ def _generate_passwords(self):
             'DATASTORE_RW_PASSWORD': generate_db_password(),
             }
 
-    def start_web(self, production=False):
+    def start_web(self, host='127.0.0.1', production=False):
```
wardi: call this address and add it to the end of the args, just in case some other code is depending on production being the first arg.

JackMc: Sounds good

wardi:
is this ready for review?

JackMc:
I'm not exactly sure how to choose between the specified host (which we can't tell if it's the default or an explicitly specified one) vs the saved host

JackMc:
The saved host is currently chosen at-create. I don't know if we want to be writing to the config at start-time

wardi:
we already do save the config if the port is set, don't we?

JackMc:
No, I don't think so

JackMc:
https://github.com/datacats/datacats/blob/master/datacats/environment.py#L489 Port is picked here then not saved

JackMc:
https://github.com/datacats/datacats/blob/master/datacats/environment.py#L536 This seems to be where the port is saved, but only on the CKAN side

wardi:
It's saved here https://github.com/datacats/datacats/blob/master/datacats/cli/manage.py#L71-L73

JackMc:
There we go :smile:

datacats/cli/create.py
```diff
@@ -20,7 +20,7 @@ def create(opts):
     """Create a new environment
 
 Usage:
-  datacats create [-bin] [--ckan=CKAN_VERSION] ENVIRONMENT_DIR [PORT]
+  datacats create [-bn] [-i|--address=IP] [--ckan=CKAN_VERSION] ENVIRONMENT_DIR [PORT]
```
wardi: I have to eat my words here. I guess it is possible for to set an address and still not start the containers.

datacats/cli/create.py
```diff
@@ -28,6 +28,7 @@ def create(opts):
   -b --bare               Bare CKAN site with no example extension
   -i --image-only         Create the environment but don't start containers
   -n --no-sysadmin        Don't prompt for an initial sysadmin user account
+  --address=IP            Address to listen on (Linux-only) [default: 127.0.0.1]
```
wardi: I've tried to sort the options alphabetically, just 'cause.

Please merge after those two changes.

Issue #86, opened 2015-05-11
Make the user experience more robust for the novice user


deniszgonjanin:
The process to getting your first environment running should be more seamless for the new user. Right now if they skip `datacats pull`, they need to jump a couple of hoops in order to recover:
```
$ pip install datacats
$ datacats create test
docker.errors.APIError: 404 Client Error: Not Found ("No such image: datacats/web:preload-2.3 (tag: preload-2.3)")
$ datacats pull
...........
$ datacats create test
Environment data directory /home/vagrant/.datacats/test already exists
$ datacats purge test
$ datacats create test
Environment directory already exists
$ datacats purge test --delete-environment
$ datacats create test
```

JackMc:
@deniszgonjanin Do you think #89 does enough to close this?

deniszgonjanin:
yes, feel free to close this once it's merged

PR #89, opened 2015-05-14
Error message when a user tries to create an environment without pull…

JackMc:
…ing down the necessary images with 'datacats pull'

Partially addresses #86.

### DIFF
--- a/datacats/docker.py
+++ b/datacats/docker.py
@@ -185,7 +185,7 @@ def image_exists(name):
     """
     # This returns a list of container dicts matching (exactly)
     # the name `name`.
-    return len(_docker.images(name=name)) >= 1
+    return bool(_docker.images(name=name))
 
 def remove_container(name, force=False):
     """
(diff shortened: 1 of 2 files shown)


JackMc:
Note that when the datacats/lessc container is introduced, this will need to be changed. Maybe we should centralize the list of needed containers to the environment.py file.

datacats/docker.py
```diff
@@ -177,6 +177,16 @@ def run_container(name, image, command=None, environment=None,
...
 
+def image_exists(name):
+    """
+    Queries Docker about if a particular image has been downloaded.
+
+    :param name: The name of the image to check for.
+    """
+    # This returns a list of container dicts matching (exactly)
+    # the name `name`.
+    return len(_docker.images(name=name)) >= 1
```
wardi: `return bool(_docker.images(name=name))` is a little more pythonic

Issue #90, opened 2015-05-14
Any datacats command hangs for a long time if boot2docker is not up


JackMc:

JackMc:
I'm working on this.

PR #91, opened 2015-05-14
Error when boot2docker isn't currently up.

JackMc:
Addresses #90

### DIFF
--- a/datacats/docker.py
+++ b/datacats/docker.py
@@ -60,17 +60,14 @@ def _get_docker():
                                                 stderr=devnull_f
                                                 ).strip()
             if status == 'poweroff':
-                # The extra newline at the start makes it look prettier during
-                # "datacats create" and related commands which use the .....
-                raise DatacatsError('''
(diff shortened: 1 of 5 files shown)


JackMc:
Need to lazy-instantiate Docker client object.

JackMc:
Sorted!

datacats/docker.py
```diff
@@ -36,19 +39,53 @@ def cmp(a, b):
...
+                # "datacats create" and related commands which use the .....
+                raise DatacatsError('''
+boot2docker is not powered on. Please run "boot2docker up".''')
+        except OSError:
+            # We're on Linux, or boot2docker isn't installed.
+            pass
+        except subprocess.CalledProcessError:
+            raise DatacatsError('''
+You have not created your boot2docker VM. Please run "boot2docker create" to do
+so.''')
```
wardi: minor inconsistency here in that the message starts with '\n'. Consider

```python
raise DatacatsError(
    'You have not created your boot2docker VM. Please run '
    '"boot2docker create" to do so')
```

PR #93, opened 2015-05-14
Rewrite cli entry point to support reporting errors with exceptions

dborzov:
We have two types of errors:
1) "user errors" where the requested action cannot be performed and the user needs to receive the explaining message on why this is the case
2) "programming errors" where we have unintended behaviour which should be fixed

The Pythonic way is to raise an Exception and let it propagate all the way up the stack. We then catch that Exception within the entry point (main.py:main) and show the detailed error message to the user.

The two types of Exceptions will correspond to the two types of errors:
1) `DataCatsError` class instances and its children for "user errors"
2) All the other exceptions will their specific types for "programming errors"

Returning 1 and 0 codes for errors leads to confusing user experience where error messages are not very helpful and we probably should move away from that.

### DIFF
--- a/datacats/cli/deploy.py
+++ b/datacats/cli/deploy.py
@@ -43,7 +43,7 @@ the environment name.
         if not profile.create(environment, target_name, stdout):
             return 1
 
-    profile.deploy(environment, target_name, stdout):
+    profile.deploy(environment, target_name, stdout)
     print "Deployed source to http://{0}.datacats.io".format(target_name)
 
     if opts['--create']:
--- a/datacats/cli/deploy.py
+++ b/datacats/cli/deploy.py
@@ -43,9 +43,7 @@ the environment name.
         if not profile.create(environment, target_name, stdout):
             return 1
 
-    if not profile.deploy(environment, target_name, stdout):
-        return 1
-
+    profile.deploy(environment, target_name, stdout):
     print "Deployed source to http://{0}.datacats.io".format(target_name)
 
     if opts['--create']:
--- a/datacats/cli/main.py
+++ b/datacats/cli/main.py
@@ -4,7 +4,6 @@
 # the terms of the GNU Affero General Public License version 3.0.
 # See LICENSE.txt or http://www.fsf.org/licensing/licenses/agpl-3.0.html
 
-
 """datacats command line interface
 
 Usage:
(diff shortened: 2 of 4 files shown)


datacats/cli/deploy.py
```diff
@@ -43,9 +43,7 @@ def deploy(environment, opts):
         if not profile.create(environment, target_name, stdout):
             return 1
 
-    if not profile.deploy(environment, target_name, stdout):
-        return 1
-
+    profile.deploy(environment, target_name, stdout):
```
JackMc: Typo? : still there

dborzov: oh, right, tnx! :)

datacats/cli/main.py
```diff
@@ -58,61 +58,73 @@
...
+          environment = Environment.load(opts['ENVIRONMENT'] or '.')
+          return command_fn(environment, opts)
+      return command_fn(opts)
     except DatacatsError as e:
         print e
-        return 1
+        sys.exit(1)
+
+
+def _parse_arguments(args):
```
wardi: indenting is off here, please use 4 spaces

datacats/cli/profile.py
```diff
@@ -15,31 +17,28 @@ def get_working_profile(environment):
...
+        if key_recognized:
+            # we sucessfully loaded the ssh key and tested it by connecting to the server, so all is good and
+            # we have got ourselves a working profile
+            return profile
+
+        user_error_message = """Your ssh key (which is an equivalent of your password so that datacats.io could recognize you) does not
+        does not seem to be recognized by the datacats.io server.\n \n
+        Most likely it is because you need to go to www.datacats.com/account/key and add the following public key: \n \n {public_key} \n \n
+        to your profile so that datacat's server could recognize you. So maybe try that?
+        If the problem persists, please contact the developer team.""".format(public_key=profile.read_public_key())
```
wardi: These messages include lots of white-space on the lines following the first. They also go past 90 or so characters making them wrap in common editor configurations. Also "does not" is repeated. Try formatting like this instead:

```python
user_error_message = ("Your ssh key (which is an equivalent of your password so that "
    "datacats.io could recognize you) does not seem to be recognized by the datacats.io server.\n\n"
    "Most likely it is because you need to go to www.datacats.com/account/key and add the "
...
```
```

**Comment by dborzov** [comment]
@wardi , fixed , please CR

File: datacats/userprofile.py [hunk 4]
```diff
@@ -86,13 +113,13 @@ def create(self, project, target_name, stream_output=None):
             web_command(
                 command=[
                     "ssh", _project_user_host(project), "create", target_name,
-                    ],
+                ],
```
wardi: hey, why are you moving my brackets and parens? either style is fine according to pep8, I tend to line them up with the indent.

dborzov: I turned on PEP8 autoformat, and it auto formatted this for me. :yum:

dborzov: :banana::banana::banana:

wardi: Go ahead and merge after those changes are removed

Issue #95, opened 2015-05-14
datacats purge is a bit too easy


deniszgonjanin:
It's trivial to nuke a datacats environment with `datacats purge`. It might be wise to make it a bit harder, adding a confirmation prompt by default, and adding a `-y` flag for when it needs to be automated.

JackMc:
:+1:  I've done this by accident before

PR #96, opened 2015-05-14
Add prompt to datacats purge by default

JackMc:
A --yes/-y option is available if purge is being automated.

### DIFF
--- a/datacats/cli/purge.py
+++ b/datacats/cli/purge.py
@@ -13,10 +13,11 @@ def purge(opts):
     """Purge environment database and uploaded files
 
 Usage:
-  datacats purge [--delete-environment] [ENVIRONMENT]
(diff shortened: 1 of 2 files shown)


JackMc:
Addresses issue #95

datacats/cli/purge.py
```diff
@@ -26,6 +27,15 @@ def purge(opts):
     except DatacatsError as e:
         environment = Environment.load(opts['ENVIRONMENT'], data_only=True)
 
+    if not opts['--yes']:
+        inp = None
+        # Nothing (default, n), y and n are our valid inputs
+        while inp not in ['y', 'n', '']:
```
wardi: `while inp.lower()[:1] not in ...`

JackMc: Smart!

Issue #15, opened 2015-01-03
Add support for less-watching container


wardi:
currently we only regenerate main.debug.css in the preload image. Add support for running the less-watching and automatic css rebuilding from a node container.

wardi:
@JackMc would you put this on your list? We're going to need it soon for rebuilding the css from less files. The current code that does this is under the docker directory

JackMc:
Sounds good! :)

JackMc:
https://github.com/jgreene/watch-less This tool does what we want to do. Should I use it or should I write my own script?

PR #97, opened 2015-05-14
Add a "datacats less" command to recompile less files.

JackMc:
Note that datacats/lessc is going to need to be pushed to work properly.


JackMc:
Addresses #15

wardi:
We have some duplication now between docker/lessc/Dockerfile and docker/compile_less.sh which is used for building the ckan preload-2.3 image, can that be factored out?

Please also add this command to the docs/commands.rst and add a circleci test.

JackMc:
Okie dokie. I didn't know that the docs/commands.rst existed! I should add the migrate there too on the other branch.

wardi:
@deniszgonjanin does this work for you?

deniszgonjanin:
@wardi yes it does

PR #104, opened 2015-05-15
Rewrite create commands with Exception handling via Python exceptions

dborzov:
More of the same. @wardi , plz CR


wardi:
looks great! :+1:

JackMc:
https://circleci.com/gh/datacats/datacats/83 CI fails after this PR due to removed "address" parameter on create_environment. Working on a fix.

JackMc:
We really need to get the indicators back onto PRs.

PR #110, opened 2015-05-19
Check for all containers running before saying we're already running

### LINKED ISSUE #79: If Database (or Solr) Container is Stopped, 'datacats start' Reports As Already Running
To reproduce:

```
datacats start
# Simulate a database failure/a machine dying
docker kill datacats_postgres_myenv
# Reports that ckan instance is already running
datacats start
```

JackMc:
Fixes #79

### DIFF
--- a/datacats/cli/manage.py
+++ b/datacats/cli/manage.py
@@ -46,9 +46,12 @@ ENVIRONMENT may be an environment name or a path to an environment directory.
 Default: '.'
 """
     environment.require_data()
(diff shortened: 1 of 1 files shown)


wardi:
looks good

wardi:
Are we doing this anywhere else, though? Maybe make a method on the Environment object to call?

JackMc:
Good idea! Reopening...

PR #116, opened 2015-05-19
Fix wording on error messages

JackMc:


wardi:
Let's use the "DataCats" spelling just for the company or the web site. The github repo and command are always "datacats"

Would you fix the indentation while you're there too?

JackMc:
Will fix after class

JackMc:
Done

PR #122, opened 2015-05-20
Add 'datacats/lessc' to required images.

JackMc:
Don't merge until datacats/lessc is on Docker Hub.

### DIFF
--- a/datacats/environment.py
+++ b/datacats/environment.py
@@ -942,7 +942,8 @@ def require_images():
     """
     if (not image_exists('datacats/web') or
             not image_exists('datacats/solr') or
-            not image_exists('datacats/postgres')):
+            not image_exists('datacats/postgres') or
+            not image_exists('datacats/lessc')):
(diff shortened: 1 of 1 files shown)


JackMc:
After some discussion, we came to the conclusion that the most friendly way to do this is to add an option (i.e. --devel) to the "datacats pull" command to pull down development tools, too.

deniszgonjanin:
Isn't the whole library development tools?

JackMc:
As in things like the lessc container, stuff that would be used in active development and not deployment.

In addition commands like 'datacats less' would pull down containers if needed. The 'postgres', 'web', and 'solr' containers would still be downloaded by default by 'pull'.

deniszgonjanin:
+1 to downloading automatically on `datacats less`

JackMc:
Closing pending A Better Way (TM)

JackMc:
Working on a new PR :)

Issue #121, opened 2015-05-20
"datacats pull" does not fail when it cannot get an image


JackMc:
```
Pulling image datacats/web....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
Pulling image datacats/web:preload-2.3..........................
Pulling image datacats/postgres.........................................................................................................................................................................................................................................................................................................
Pulling image datacats/solr..................................................................................................................................................................................................................................................................................................................................
Pulling image datacats/lessc.{
  "errorDetail": {
    "message": "Error: image datacats/lessc: not found"
  }, 
  "error": "Error: image datacats/lessc: not found"
}
```

Return code: 0

JackMc:
Fixed by #123

PR #123, opened 2015-05-20
Make "datacats pull" fail if it can't download an image.

JackMc:


JackMc:
Addresses #121

wardi:
go ahead and merge :+1:

PR #124, opened 2015-05-20
Automatically download datacats/lessc as needed

JackMc:
Also allow specifying a --devel option to 'datacats pull' to download all images.

### DIFF
--- a/datacats/cli/less.py
+++ b/datacats/cli/less.py
@@ -1,7 +1,6 @@
-import sys
+from datacats.docker import image_exists
 
-from datacats.docker import image_exists, pull_stream
-from datacats.error import DatacatsError
+from datacats.cli.pull import pull_image
 
 
 LESSC_IMAGE = 'datacats/lessc'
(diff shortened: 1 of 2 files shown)


datacats/cli/pull.py
```diff
@@ -22,9 +25,16 @@ def pull(opts):
     """Download or update required datacats docker images
 
 Usage:
-  datacats pull
+  datacats pull [-d]
+
+Options:
+  -d --devel         Pull extra "development" images. These
```
wardi: I like `--all`
"Pull optional images as well as required images. Optional images will be pulled when needed, ..."

JackMc: Okay :)

datacats/cli/less.py
```diff
@@ -7,5 +16,17 @@ def less(environment, opts):
...
+        for status in pull_stream(LESSC_IMAGE):
+            if 'status' not in status:
+                raise DatacatsError('Docker error: ' +
+                        (status['error'] if 'error' in status else 'Unknown error.'))
+            else:
+                sys.stdout.write('.')
+                sys.stdout.flush()
+        sys.stdout.write('\n')
+        sys.stdout.flush()
+
```
wardi: don't duplicate this code, make a function that the pull command uses and call it instead

dborzov:
Woah, I had no idea datacats even wraps less parser.

Lazy loading of `datacats/less` looks neat though!  :+1:

Issue #49, opened 2015-03-23
test for ssh connectivity


deniszgonjanin:
`datacats deploy` may fail because the user is behind a firewall, or their VM being wonky, or any other reason. datacats should return a different error when a connection cannot be established versus when it cannot authenticate successfully with the deployment environment

dborzov:
Fixed

PR #126, opened 2015-05-20
Show right message when 'ssh test' is not able to connect to the server

dborzov:
This addresses [#49](https://github.com/datacats/datacats/issues/49)


JackMc:
:+1:

Issue #64, opened 2015-04-15
When things go wrong, path to recovery is often unclear


deniszgonjanin:
For example, a boot2docker VM dies. Upon recovery, doing `datacats start` on a project will show:
```
Traceback (most recent call last):
  File "/Users/dz/Envs/datacats/bin/datacats", line 9, in <module>
    load_entry_point('datacats==0.9dev1', 'console_scripts', 'datacats')()
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/datacats-0.9dev1-py2.7.egg/datacats/cli/main.py", line 115, in main
    return command_fn(project, opts)
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/datacats-0.9dev1-py2.7.egg/datacats/cli/manage.py", line 54, in start
    reload_(project, opts)
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/datacats-0.9dev1-py2.7.egg/datacats/cli/manage.py", line 80, in reload_
    project.start_web(opts['--production'])
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/datacats-0.9dev1-py2.7.egg/datacats/project.py", line 502, in start_web
    self._run_web_container(port, command)
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/datacats-0.9dev1-py2.7.egg/datacats/project.py", line 571, in _run_web_container
    5000: port if is_boot2docker() else ('127.0.0.1', port)},
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/datacats-0.9dev1-py2.7.egg/datacats/docker.py", line 157, in run_container
    port_bindings=port_bindings)
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/docker_py-1.1.0-py2.7.egg/docker/client.py", line 904, in start
    self._raise_for_status(res)
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/docker_py-1.1.0-py2.7.egg/docker/client.py", line 119, in _raise_for_status
    raise errors.APIError(e, response, explanation=explanation)
docker.errors.APIError: 406 Client Error: Not Acceptable ("Cannot start container 4e5e47801f4201c0365ce5b363263e27e7576480259289f7f7bbc8a84867932e: container datacats_venv_donneesqc not found, impossible to mount its volumes")
```

The correct path to recovery is to do:
```
datacats purge
datacats init
```

[...]

PR #130, opened 2015-05-21
Add extra checks for volume containers existence

JackMc:
Addresses (partially) #64


JackMc:
This is pretty small and ready for CR :smile:

datacats/environment.py
```diff
@@ -301,6 +310,9 @@ def require_data(self):
             raise DatacatsError('Environment datadir damaged. '
                                 'Try "datacats purge" followed by'
                                 ' "datacats init".')
+        if not self.volumes_exist():
+            raise DatacatsError('Volumes damaged. '
+                                'Try "datacats purge" followed by "datacats init"')
```
wardi: A little friendlier and more cautious, like:
`Volume containers could not be found. To reset and discard all data use "datacats purge" followed by "datacats init"`

JackMc: Okie, that error message was copied from the original issue :stuck_out_tongue:

Issue #128, opened 2015-05-21
Container names are hardcoded when referred to


JackMc:
This should be factored out into a method.

PR #131, opened 2015-05-21
Abstract container naming into a method

JackMc:
This addresses #128

### DIFF
--- a/datacats/environment.py
+++ b/datacats/environment.py
@@ -317,7 +317,7 @@ class Environment(object):
         Populate venv directory from preloaded image
         """
         if is_boot2docker():
-            data_only_container('datacats_venv_' + self.name,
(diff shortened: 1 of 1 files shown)


JackMc:
@wardi @dborzov This is ready for CR.

dborzov:
Neat! 

Are the advantages to having a dedicated method over, say, doing this at the Environment initialization? 

My naive approach would have been to do something like

```python
  __init__(...):
       self.container = {key: "datacats_{key}_{name} '.format(key=key, name=self.name) for key in ['web', 'solr' ...]}
```

JackMc:
I decided on this because in #82 we name the venv like datacats_venv_myenv and everything else like datacats_postgres_myenv_primary

Issue #46, opened 2015-03-17
deploy to any CKAN hosting provider


wardi:
There is the beginning of this feature in a "deploy" configuration setting in .datacats-environment . The completed feature should allow a complete config section like

```ini
[deploy]
target = my.host.example.com
user = myself
ssh_host_key = my.host.example.com ssh-rsa AAAAB3NzaC....
```

Then `datacats deploy` will rsync to the specified target instead of datacats.com. The ssh_host_key is used to verify that it's talking to the correct server (found in /etc/ssh/ssh_host_key_rsa.pub, or by running ssh-keyscan)

deniszgonjanin:
Related to #98

wardi:
closing in favour of #98

PR #132, opened 2015-05-21
Refactor commands on datacats projects as methods of  datacats projects, not UserProfile

dborzov:
Let's refactor the UserProfile/Environment relationship a little bit so that it would make more sense:

* The UserProfile is currently loaded within `cmd/profile.py:get_working_profile` which is obviously not the place for it (it is not a tty subcommand). I think the UserProfile should be loaded when the environment is loaded and initialized so that it is dependency-injected into environment to `environment.profile`. See this PR for this.
* When we perform actions on the environment, we should make it an environment method, not userProfile method with environment passed as the first parameter. For example it should be `Environment.deploy`, not `UserProfile.deploy(environment)`.
* I think we should come up with a consistent nomenclature for datacats environment/project/catalog thingy once and for all and stick to it. It is confusing even to me and I work on this full time. Imagine how confused the users would be. I think **Environment** is not a good term and vote for **project** or **catalog**. What do you guys think?

What do you think: @wardi , @JackMc , @deniszgonjanin ?


wardi:
"project" was the old name for the environment, and as you've seen it still exists in the UserProfile code and inside the containers as the "/project/" directory where everything is mounted.

dborzov:
@wardi, so we are going to call it environment from now on?

wardi:
Currently we have only one profile per user. If we add support for deploying to multiple targets that might need to change, so making profile an attribute of the Environment doesn't seem right

JackMc:
For some reason, I think /project/ makes more sense in the context of the mounts in the containers. It's more intuitive to me.

wardi:
@deniszgonjanin what was the reason you liked "environment" better than "project".

I'm finding myself saying "oh, not the Python virtualenv, the datacats environment" and "oh, not the environment variables, the datacats environment" and sort of regretting the change.

dborzov:
Let's call it a datacat :) One datacat, two datacats.

JackMc:
:laughing:

JackMc:
But yeah, it's really confusing to talk about the project dir vs datadir in terms of the word 'environment'.

deniszgonjanin:
I forget exactly, but we did have a long discussion about this. Environment sucks because of virtualenv, but it more properly describes what datacats is doing - creating self-contained ckan development environments. Not dissimilar to virtualenv.

Project sucks because inside every datacats 'project' you have a bunch of other projects. And a datacats 'project' is not really a project in itself, it's more of a collection of different projects packaged together for devops purposes - an "environment"

wardi:
datadir is a lousy name too :-) And not only because I forgot to put a _ in there. Naming is hard.

datacats/environment.py
```diff
@@ -57,6 +62,13 @@ def __init__(self, name, target, datadir, ckan_version=None, port=None,
         self.deploy_target = deploy_target
         self.site_url = site_url
         self.always_prod = always_prod
+        self.profile = UserProfile()
+        if not (self.profile.ssh_private_key is None) and \
+            exists(self.profile.ssh_private_key) and \
+            exists(self.profile.ssh_public_key):
+            self.profile.test_ssh_key(self)
```
JackMc: This shouldn't always be done. Only when using a deploy-related or remote command.

wardi:
yes ckan is referred to as a "project". I think everything else in there is a "package" or "extension".

deniszgonjanin:
Yea, so I think because datacats is primarily a devops tool, "environment" is more suitable. But you can certainly paint this shed many colors

datacats/environment.py
dborzov: Good point, thanks :)

dborzov:
Implements [#46](https://github.com/datacats/datacats/issues/46) with the following `.datacats_environment` options to overwrite the default remote server settings :
```
[deploy]
remote_server = 192.168.0.13
remote_server_user = borzov
remote_server_key = 192.168.0.13 ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQDpH6rYXkR5bqKHDRdZN4enLJC0I3b039SH3rl4siJQlgRCffB1VOe1Sz1Bl6/f5kGglRXvLMV49uslUEh5ej0r0YY7Io6+2BOpA7a/SWLWFXM/eyKIVA917VK8zeHZczwf91bl0P3Vw4vmNWei4wvd6kBKR3wNOppEM4aAb6w17YjehR6F1xSagnbXsIcvrDSe73ym/7RJtcJM30rA8vH3WWsPpuZf1G2/TZAmi0ky44zDwAFBPj+nmUBIG8dhAW2UaLaKttD7VLbk7NKo6Okq6nnT4q++BOEmbq6qpspabfTJp7PZIURhx8ygvU5I4AtE0rF6B0qO2iEabY/VPiRB
``` 

Also, this PR refactors and cleans the pertinent things:
1) deletes `datacats/userprofile.py:project_user_host`
2) deletes `datacats/cli/profile.py` and moves that logic to `UserProfile` constructor
3) moves the repetitive docker's remote connection-related binds to a dedicated method

@wardi @JackMc plz CR :)

datacats/environment.py
```diff
@@ -219,10 +222,20 @@ def load(cls, environment_name=None, data_only=False):
         except NoOptionError:
             always_prod = False
         try:
-            deploy_target = cp.get('deploy', 'target', None)
-        except NoSectionError:
-            deploy_target = None
+            deploy_target = cp.get(
+                'deploy', 'remote_server_user', None) \
+                + "@" + cp.get('deploy', 'remote_server', None)
+        except NoOptionError:
```
JackMc: Should also except NoSectionError, as a lack of a [deploy] section is the default behaviour.

datacats/cli/main.py
```diff
@@ -70,13 +76,25 @@ def main():
...
+        # we load UserProfile and test our communication
+        user_profile = UserProfile()
+        user_profile.test_ssh_key(environment)
+
+        return command_fn(environment, opts, profile=user_profile)
+
     except DatacatsError as e:
         if sys.stdout.isatty():
-            # error message to have colors if stdout goes to shell
+                # error message to have colors if stdout goes to shell
```
JackMc: This indentation doesn't make sense

datacats/environment.py
```diff
@@ -889,6 +903,21 @@ def compile_less(self):
             ro={COMPILE_LESS: '/project/compile_less.sh'})
         remove_container(c)
 
+    def remote_command_binds(self, profile, include_project_dir=False):
+        known_hosts = self.datadir + \
+            '/known_hosts' if exists(self.datadir +
+            '/known_hosts') else KNOWN_HOSTS
+        # import ipdb
+        # ipdb.set_trace()
```
JackMc: pdb

dborzov: tnx :)

dborzov:
@wardi , @JackMc all's fixed :)

JackMc:
:+1: Looks good :smile:

datacats/cli/deploy.py
```diff
@@ -28,8 +27,8 @@ def deploy(environment, opts):
 TARGET_NAME is the name of the environment on DataCats.com. Defaults to
 the environment name.
 """
-    profile = get_working_profile(environment)
-
+    if not profile:
+        raise Exception("Need to pass a loaded user_profile instance")
```
wardi: In general it's bad to raise Exception. Wouldn't this fire if the above function call fails, not if user_profile wasn't passed?  If this will only happen if we did something wrong, I wouldn't bother checking here, just let it fail later on instead.

wardi: Sorry, I misread, that's the part you removed. My comment about the exception stands though

dborzov: Yeah, it is pretty silly this way.

datacats/environment.py
```diff
@@ -218,10 +221,27 @@ def load(cls, environment_name=None, data_only=False):
...
+        # we overwrite the default datacats.com one
         try:
-            deploy_target = cp.get('deploy', 'target', None)
-        except NoSectionError:
-            deploy_target = None
+            deploy_target = cp.get(
+                'deploy', 'remote_server_user', None) \
+                + "@" + cp.get('deploy', 'remote_server', None)
+        except:
+            deploy_target = DEFAULT_REMOTE_SERVER_TARGET
```
wardi: Avoid bare excepts that don't also "raise". This style is better, and the same for the change below:

```python
try:
    deploy_target = (cp.get('deploy', 'remote_server_user')
        + '@' + cp.get('deploy', 'remote_server'))
except (NoOptionError, NoSectionError):
    deploy_target = DEFAULT_REMOTE_SERVER_TARGET
```
```

Comment (dborzov) [inline datacats/environment.py]: fixed

File: datacats/environment.py [hunk 7]
```diff
@@ -218,10 +221,27 @@ def load(cls, environment_name=None, data_only=False):
...
+
+        # if remote_server's ssh public key is given,
+        # we overwrite the default datacats.com one
+        try:
+            known_hosts_path = datadir + '/known_hosts'
```
wardi: I've been sticking these sorts of temporary files in the "run" directory, although that's not perfect. Maybe we should be using proper temp files

JackMc: tempfile is really easy to use and multi-platform.

datacats/environment.py
```diff
@@ -218,10 +221,27 @@ def load(cls, environment_name=None, data_only=False):
...
+        except (NoOptionError, NoSectionError):
+            deploy_target = DEFAULT_REMOTE_SERVER_TARGET
+
+        # if remote_server's ssh public key is given,
+        # we overwrite the default datacats.com one
+        try:
+            known_hosts_path = datadir + '/known_hosts'
+            with open(known_hosts_path, "wb") as known_hosts:
+                known_hosts.write(cp.get(
+                    'deploy', 'remote_server_key', None))
```
wardi: remove the `,None` or the correct exception won't get raised

datacats/environment.py
```diff
@@ -218,10 +221,27 @@ def load(cls, environment_name=None, data_only=False):
...
+        # if remote_server's custom ssh connection
+        # address is defined,
+        # we overwrite the default datacats.com one
         try:
-            deploy_target = cp.get('deploy', 'target', None)
-        except NoSectionError:
-            deploy_target = None
+            deploy_target = cp.get(
+                'deploy', 'remote_server_user', None) \
+                + "@" + cp.get('deploy', 'remote_server', None)
```
wardi: remove the `,None`s here too

wardi:
the rest is ok. Feel free to s/project/environment/ in the method parameters if you like (or leave it for later)

datacats/environment.py
```diff
@@ -218,10 +221,27 @@ def load(cls, environment_name=None, data_only=False):
...
+        except (NoOptionError, NoSectionError):
+            deploy_target = DEFAULT_REMOTE_SERVER_TARGET
+
+        # if remote_server's ssh public key is given,
+        # we overwrite the default datacats.com one
+        try:
+            known_hosts_path = datadir + '/known_hosts'
+            with open(known_hosts_path, "wb") as known_hosts:
```
JackMc: This doesn't handle the known_hosts file not existing.

dborzov:
Ok, fixed. Plz CR :dolphin: :dolphin:

wardi:
:metal:

dborzov:
Yay!!

PR #134, opened 2015-05-29
Let's get ourselves a pretty documentation badge!

dborzov:
The one from ReadTheDocs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # datacats
 
 [![Circle CI](https://circleci.com/gh/datacats/datacats.svg?style=svg)](https://circleci.com/gh/boxkite/datacats)
+[![docs](https://readthedocs.org/projects/docs/badge/?version=latest)](http://docs.datacats.com/)
+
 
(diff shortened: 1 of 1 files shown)


wardi:
We're generating our own docs, not using RTD. If the license on the image allows let's host the image ourselves

dborzov:
Hm, I am not sure how to check the license for this image. 

What if we had the docs built on RTD (so the intended use is legally justified) but made the badge link to our self-hosted documentation? The account is free for the open source projects.

dborzov:
BTW, it turns out Github Readme parser supports reSturcturedText for Readmes just like Markdown! 

Here is an example: https://github.com/rtfd/readthedocs.org/blob/master/README.rst

JackMc:
Hey Dima can you do a rerun on the CI build? There's no reason I should be failing...

dborzov:
on it

JackMc:
ReadTheDocs seems to be licensed under MIT, but no mention of image licensing.

JackMc:
Also MIT's wording only covers "software" and "associated documentation"

PR #139, opened 2015-06-01
Add some unit tests for environments

JackMc:


JackMc:
This is just for CR, don't merge yet.

dborzov:
Whoah some actual tests!! Kudos

wardi:
I don't like unit tests that modify things in the users home directory.

Instead let's add parameters to Enviroment methods to allow them to be unit tested. These  parameters will default to the normal things they need to call (datacats/docker functions etc.) but in our tests we can pass in closures that return reasonable values and nothing gets modified on the host.

e.g.
```python

def test_start_web_creates_run_ini_properly():
    ...
    env.start_web(
        _is_boot2docker=lambda: False,
        _exists=lambda x: True,
        _isdir=lambda x: True,
        _SafeConfigParser=MockSafeConfigParser,
        _open=mock_open)
    # verify contents of run_ini written to mock_open "file"
    ...
```

JackMc:
That's a great idea! A lot of parameters to pass though. What do you think about making those constructor parameters and then making the methods members of the object?

wardi:
Like you suggested we could change the way we import these functions we call so that we can mock them in our tests (import the module then call with modulename.) . That keeps the testing cruft in our code to a minimum and is only slightly magical. We can use some common mocking library to make it even more acceptable.

JackMc:
Okay, I'm going to look into mock frameworks after this.

wardi:
closing in favour of building tests based on #213

Issue #107, opened 2015-05-17
Configuration options plan


wardi:
1. **global**: ~/.datacats/config
2. **user**: ~/.datacats/user-profile/config
3. **env**: *path-to-environment-dir*/.datacats-environment
4. **datadir**: ~/.datacats/<name>/

where a setting may appear in multiple locations the higher-numbered setting overrides

| option | global | user | env | datadir |
| --- | --- | --- | --- | --- |
| DB passwords | | | | :white_check_mark: |
| site_url | | | | :white_check_mark: |
| port | | | :white_check_mark: | :white_check_mark: |
| address | :white_check_mark: | | | :white_check_mark: |
| deploy_target | | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| ckan_version | | | :white_check_mark: | |
| future service containers | | | :white_check_mark: | |
| ssh_public_key | | :white_check_mark: | |
| ssh_private_key | | :white_check_mark: | |
| DNS | :white_check_mark: | | | |
| always_prod | :white_check_mark: | | | |

This is how I think our settings should be arranged. These are not yet in the right places:
- [ ] site_url
- [ ] port
- [ ] address
- [ ] deploy_target
- [ ] DNS
- [ ] always_prod

JackMc:
As some of these are going to be breaking changes (i.e. moving always prod to only global will break some people's confs) should we still look in the old location?

JackMc:
Should a global configuration option be set when it's changed on the command line, or just overridden for that one run?

wardi:
good question. Probably just that one run, but that's inconsistent with what we're doing for the environment settings like port and address.

how about we document our slightly uncommon saving-behaviour of the port and address options and do the same for dns, just tell the user it will be saved and applied to all new and existing environments.

It's either that or tell users where to find the file to edit to make the change permanent. That option seems harder for us and them.

JackMc:
Would it be better just to open a file to each of these configuration files at constructor-time and then deal with them as needed in their relevant methods? Currently, we read and delete them as needed which might get messy if as suggested we have 4 config files to read from

wardi:
This is already out of date (since multisite) and I'd like to revisit some of these since we're also not storing the --production and --no-reload options (and clobbering them on datacats install) We should be more consistent and predictable about how we handle options

JackMc:
Update for sites

PR #141, opened 2015-06-02
Add syslog option to all commands which start_web

JackMc:


JackMc:
This addresses #141

wardi:
@deniszgonjanin does this work for you? Does it log to syslog on the host or in the containers?

deniszgonjanin:
I am getting an error on this branch with any datacats command:

```
datacats reload
  File "/Users/dz/Envs/datacats/bin/datacats", line 9, in <module>
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/pkg_resources.py", line 353, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/pkg_resources.py", line 2321, in load_entry_point
    return ep.load()
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/pkg_resources.py", line 2048, in load
    entry = __import__(self.module_name, globals(),globals(), ['__name__'])
  File "/Users/dz/source/datacats/datacats/cli/main.py", line 39, in <module>
    from datacats.cli import create, manage, install, pull, purge, shell, deploy, less
  File "/Users/dz/source/datacats/datacats/cli/create.py", line 11, in <module>
    from datacats.environment import Environment
  File "/Users/dz/source/datacats/datacats/environment.py", line 22, in <module>
    from datacats.docker import (web_command, run_container, remove_container,
  File "/Users/dz/source/datacats/datacats/docker.py", line 30, in <module>
    from docker.utils import kwargs_from_env, compare_version, create_host_config, LogConfig
ImportError: cannot import name LogConfig
```

Does the `docker-py` requirement need to be updated?

JackMc:
Looking into it :p

JackMc:
Could try the install and/or develop again. It works fine for me.

I do need to enable syslog on the web container, though. Need to work more on this.

deniszgonjanin:
@JackMc what version of `docker-py` do you have?

JackMc:
1.2.2

deniszgonjanin:
The error disappears on 1.2.2. setup.py should be updated in this PR with docker-py 1.2.2 set as the minimum version

deniszgonjanin:
@wardi The --syslog option doesn't seem useful on OSX, but this should still be useful for production purposes

wardi:
@deniszgonjanin we have a force_production setting for production, this seems related, maybe we should handle them the same way?

JackMc:
Do you mean have all production systems log to syslog?

deniszgonjanin:
Well, the production site I have has logs that are unmanageable after only a couple of weeks. So maybe yes, syslog always on in production mode.

wardi:
Ok, merging as-is. We'll probably need to add a force_syslog option later but that can wait until we fix our configuration file locations #107

PR #142, opened 2015-06-02
Add a little test to identify issues with docker connection

dborzov:
This little test blows up when connection to the docker is broken. Which can be an indicator of issues like:
1) test environment not being set properly
2)  [this bug](https://github.com/docker/docker-py/issues/465) showing up again

### DIFF
--- a/datacats/tests/test_docker.py
+++ b/datacats/tests/test_docker.py
@@ -1,0 +1,25 @@
+from unittest import TestCase
+from docker.utils import kwargs_from_env
+from docker.client import Client
+
+
+class TestOpenSSLBug(TestCase):
+    """
+      Tests if communication between Docker and datacats fails
+      (There could be various reasons for that like:
+         1) Docker not installed properly
+         2) Docker's envvariable not set up
+         3) that weird bug one gets on MacOSX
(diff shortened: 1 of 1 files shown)


JackMc:
I think that having this run as a check at runtime might be a better idea, no?

I know that as a user things like CONNECTION REFUSED would freak me out.

JackMc:
Maybe put this check in _get_docker in docker.py?

JackMc:
(CI failure can be fixed by specifying Docker version to Client constructor.

dborzov:
Good idea! We should do this and maybe refer people to updated (FAQ)[http://docs.datacats.com/faq.html#insecureplatformwarning] or something

wardi:
@dborzov CI is still broken, please fix as suggested in https://github.com/datacats/datacats/pull/142#issuecomment-108039667

JackMc:
Don't think this is really relevant anymore

wardi:
Doesn't this produce a clear failure when something as basic as the connection to docker is broken?

JackMc:
I mean but this is at test-time not at run-time

wardi:
That's true. closing

Issue #147, opened 2015-06-03
datacats create should work without internet connection


wardi:
Everything required is in the preloaded virtualenv, we shouldn't fail from datacats create if we can't reach pypi.

wardi:
@JackMc  This is in the wrong place: https://github.com/datacats/datacats/blob/master/datacats/scripts/install_package.sh#L11  pip install -e doesn't download anything

Please move it to install_requirements.sh instead

JackMc:
Okay.

PR #148, opened 2015-06-03
Move wget check to a more appropriate location.

JackMc:


JackMc:
Addresses #147

JackMc:
Closing due to better fix in #150

PR #149, opened 2015-06-03
Pass all  keyed arguments from environment.new to environment-__init__

dborzov:
For the greater good!

### DIFF
--- a/datacats/environment.py
+++ b/datacats/environment.py
@@ -107,7 +107,7 @@ class Environment(object):
             pdir.write(self.target)
 
     @classmethod
-    def new(cls, path, ckan_version, port=None, address=None):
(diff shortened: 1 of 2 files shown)


datacats/environment.py
```diff
@@ -107,7 +107,7 @@ def _update_saved_project_dir(self):
             pdir.write(self.target)
 
     @classmethod
-    def new(cls, path, ckan_version, port=None, address=None):
+    def new(cls, path, ckan_version, **kwargs):
```
wardi: ok, but do all callers use keyword args for port and address?

dborzov: Oh, right! I checked and fixed all such instances (I found only one).

PR #150, opened 2015-06-03
Check DNS before install in a dedicated script.

JackMc:
Addresses #147


JackMc:
Tests pass now....

wardi:
looks good. Maybe rename to check_connectivity because you're actually hitting pypi's web server

Issue #146, opened 2015-06-03
Misleading error message in WebCommandError


wardi:
@dborzov db70e469 added "SSH command to remote server failed" but WebCommand is used for every command run in the web container, not just issuing remote commands

dborzov:
Oh you are right, this needs to be fixed

JackMc:
Does #154 work for this?

PR #154, opened 2015-06-04
Fix WebCommandError remote woes

JackMc:
For some reason I like the word 'woes'.

Addresses #146

### DIFF
--- a/datacats/docker.py
+++ b/datacats/docker.py
@@ -163,7 +163,7 @@ def binds_to_volumes(volumes):
 
 def web_command(command, ro=None, rw=None, links=None,
                 image='datacats/web', volumes_from=None, commit=False,
-                clean_up=False, stream_output=None):
+                clean_up=False, stream_output=None, ssh=False):
     """
     Run a single command in a web image optionally preloaded with the ckan
     source and virtual envrionment.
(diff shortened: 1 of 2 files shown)


datacats/docker.py
```diff
@@ -103,14 +103,15 @@ def _get_docker():
 
 class WebCommandError(Exception):
 
-    def __init__(self, command, container_id, logs):
+    def __init__(self, command, container_id, logs, ssh=False):
```
dborzov: so you are suggesting using this ssh argument as a flag to differentiate between the remotecommand-specific problems and all others?

JackMc: Yes, I couldn't find out a way to differentiate them though

That's what I need feedback on :)

JackMc: Is the only place where it generates the SSH key?

dborzov: I think that that approach can get really confusing after a while. The argument name is not very telling and will need a docstring/comment at least. Another thing is that it we differentiate between classes of errors with dedicated Exception classes elsewhere and here we differentiate between the two classes with True | False argument values.

I am voting for creating a dedicated type of exception for these issues:
```
class RemoteCommandError(WebCommandError):
      ... 
```
I think it is the pythonic way.

JackMc: The problem there is that then we need some way to tell if the current command is remote-related.. 

I agree it's more pythonic though

dborzov: Good point. Do you mind if I address this? I will just go through all the occurances manually.

JackMc: Sure, go ahead. :) 

I'll close this one

dborzov: Danke shon :blush:

Issue #152, opened 2015-06-03
show log output instead of telling user to run `datacats logs`


wardi:
super common when working through bugs that prevent ckan from starting. Fix for start, reload, install in cases where it's waiting for a successful start

wardi:
fixed via e1dd1471c5

PR #156, opened 2015-06-04
Give error with logs for web container failure

JackMc:
Addresses #152

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # datacats
 
-[![Circle CI](https://circleci.com/gh/datacats/datacats.svg?style=svg)](https://circleci.com/gh/boxkite/datacats)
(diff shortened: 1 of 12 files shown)


wardi:
I'm super-excited to try this tomorrow! I think we've got a ton of little useful changes now. I'll roll out a  0.11.0 release then merge the multisite changes to master.

JackMc:
Changed over to the simple no-follow behaviour. This has the problem that it may not grab the tail-end of the logs but it was a super long wait to wait for the follow to end

wardi:
Hmm. Are there some docker options that would make this behave better (like the -it options on the command line)

wardi:
@JackMc oh. I didn't see you closed this. Should I not have merged this?

JackMc:
Yeah I dunno how it got closed

JackMc:
It looks like it merged though

PR #157, opened 2015-06-04
Figuring out our error handling

dborzov:
This
1) Addresses and fixes [#146](https://github.com/datacats/datacats/issues/146)
2) Moves all the custom error declarations reused in more than one file to `errors.py`

### DIFF
--- a/datacats/error.py
+++ b/datacats/error.py
@@ -35,7 +35,7 @@ class DatacatsError(Exception):
 
 
 class WebCommandError(Exception):
-    user_description = "Docker container "/web" command failed"
+    user_description = "Docker container \"/web\" command failed"
 
     def __init__(self, command, container_id, logs):
         self.command = command
--- a/datacats/error.py
+++ b/datacats/error.py
@@ -2,14 +2,20 @@ from clint.textui import colored
 
 
 class DatacatsError(Exception):
-
     def __init__(self, message, format_args=(), parent_exception=None):
         self.message = message
-        if parent_exception:
-            self.message += '\n\n' + '~' * 30 + \
-                "\nTechnical Details:\n" + \
-                parent_exception.__str__() + \
-                '~' * 30 + '\n'
+        if parent_exception and hasattr(parent_exception, 'user_description'):
+            vals = {
+              "original": self.message,
+              "type_description": parent_exception.user_description,
+              "message": parent_exception.__str__(),
+            }
+            self.message = "".join(["{original}\n\n",
+                            "~" * 30,
+                            "\n{type_description}:\n",
+                            "{message}",
+                            "~" * 30, "\n"]).format(**vals)
+
         self.format_args = format_args
         super(DatacatsError, self).__init__(message, format_args)
 
@@ -28,33 +34,18 @@ class DatacatsError(Exception):
 
 
 class WebCommandError(Exception):
+    user_description = '\nDocker container "/web" command failed\n'
 
     def __init__(self, command, container_id, logs):
         self.command = command
-        self.container_id = container_id
         self.logs = logs
 
     def __str__(self):
-        return \
-            ('\nDocker container "/web" command failed\n'
-             '    Command: {0}\n'
-             '    Docker Error Log:\n'
-             '    {1}\n'
-             ).format(" ".join(self.command), self.logs, self.container_id)
-
-
-class RemoteCommandError(WebCommandError):
-    def __init__(self, base_WebCommandError):
-        self.__dict__ = base_WebCommandError.__dict__
-
-    def __str__(self):
-        return \
-            ('\nSending a command to remote server failed\n'
-             '    Command: {0}\n'
-             '    Docker Error Log:\n'
-             '    {1}\n'
-             ).format(" ".join(self.command), self.logs, self.container_id)
+        return ('    Command: {0}\n'
+                 '    Docker Error Log:\n'
+                 '    {1}\n'
+             ).format(" ".join(self.command), self.logs)
 
 
 class PortAllocatedError(Exception):
-    pass
+    user_description = "Unable to allocate port"
--- a/datacats/error.py
+++ b/datacats/error.py
@@ -35,7 +35,7 @@ class DatacatsError(Exception):
 
 
 class WebCommandError(Exception):
-    user_description = '\nDocker container "/web" command failed\n'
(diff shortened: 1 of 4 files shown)


datacats/error.py
```diff
@@ -25,3 +25,36 @@ def pretty_print(self):
...
+            ('\nDocker container "/web" command failed\n'
+             '    Command: {0}\n'
+             '    Docker Error Log:\n'
+             '    {1}\n'
+             ).format(" ".join(self.command), self.logs, self.container_id)
+
+
+class RemoteCommandError(WebCommandError):
+    def __init__(self, base_WebCommandError):
+        self.__dict__ = base_WebCommandError.__dict__
```
JackMc: Cool snippet! Won't this override your __str__ implementation though?

wardi: Hmm. why is this magic required?

wardi: Oh. an exception that takes another exception as its required parameter. I don't approve :-)

Explicit is better than implicit. just pass the arguments in when you're changing the type of the exception

JackMc: Only problem is adding extra params to the webcommand constructor will be harder than a simple grep-and-replace

wardi: "where in the heck did my RemoteCommandError.logs value come from" is harder to understand with this code. Also, it's just weird to have to pass an exception to another exception's __init__.

Does the __dict__ copying prevent losing the traceback from the original exception?

Thinking about this more, why don't we pass the exception class to use to web_command (defaulting to WebCommandError) instead?

dborzov: @JackMc , fortunately `__dict__` does not include methods. This is not some godless javascriptses :)

But yeah, you guys are right it is kind of ugly this way. RemoteCommandError is a subclass of WebCommandError and the two classes have the same signature. All I would like to do here is to change that class type to subclass. Not sure how to express that.

I am not convinced if passing the exception class to the function is justified. That also would look obstructified to me.

wardi: @dborzov the alternatives are worse: raising a different exception possibly losing the useful traceback or assigning to the type of the object (just terrible) and re-raising. This is just a python limitation, hacking exceptions on the way by kind of sucks.

dborzov: @wardi reassigning `__class__` value would fix this, how do you feel about that? ;)

wardi: That was the option I described as "just terrible"

dborzov: Whoah, there is an "exception chaining" in Python 3! See this [post](http://stackoverflow.com/questions/696047/re-raising-exceptions-with-a-different-type-and-message-preserving-existing-inf)

dborzov:
Ok, @wardi @JackMc check out this approach.

No more custom RemoteCommandError classes. Custom exceptions we raise can have a user_description field, which signals to the DatacatsError catcher that their message should be shown to the user.

When we raise a DatacatsError instance we pass to it:
      1) a very generic user-oriented error message (this defines the level where we reraise the exception)
      2) the actual exception with the traceback and description as
      a parent_exception parameter.

If the parent_exception implements the basic interface of having
a _user_description string attribute, that would serve as a signal that
it also needs to be shown to the user.

So we get error message structured like this:

---------------------------------
datacats: problem was encountered
    {DatacatsError.message}
    ~~~~~~~~~~~~~~~~~~~~~~
    {parent_exception._user_description }:
       {parent_exception.message}
--------------------------------

JackMc:
:+1:

JackMc:
Looks great!!

datacats/error.py
```diff
@@ -23,5 +30,23 @@ def pretty_print(self):
         print colored.blue("-" * 40)
         print colored.red("datacats: problem was encountered:")
         for line in self.message.format(*self.format_args).split('\n'):
-            print "  ", line
+            print "    ", line
         print colored.blue("-" * 40)
+
+
+class WebCommandError(Exception):
+    user_description = '\nDocker container "/web" command failed\n'
```
wardi: \n's around the string looks like formatting mixed with content

datacats/userprofile.py
```diff
@@ -131,7 +131,6 @@ def create(self, environment, target_name, stream_output=None):
         remote_server_command(
             ["ssh", environment.deploy_target, "create", target_name],
             environment, self,
-            stream_output=stream_output,
```
wardi: why are these being removed?

dborzov: I thought we agreed that `remotecommand` is only going to return a JSON object that specifies the result in response.
Like:
```json
{
 "status": fail,
 "message": "A site with such a name already exists"
}
```
and we would not want to just print everything we receive to the user. What are the cases for which you want it left?

datacats/error.py
dborzov: Oh right, tnx

JackMc:
Latest hotfix needs to be manually merged but all it is is the fact that one of the lines that is removed in this exists in a new one

datacats/error.py
```diff
@@ -23,5 +30,23 @@ def pretty_print(self):
         print colored.blue("-" * 40)
         print colored.red("datacats: problem was encountered:")
         for line in self.message.format(*self.format_args).split('\n'):
-            print "  ", line
+            print "    ", line
         print colored.blue("-" * 40)
+
+
+class WebCommandError(Exception):
+    user_description = "Docker container "/web" command failed"
```
JackMc: Quotation needs to be alternating (either "Docker container '/web' command failed" or 'Docker container "/web" command failed'

datacats/userprofile.py
wardi: Right, so remove the parameters too?

dborzov: Should I? I guess I may do that. There is no rush though, maybe we will need it for testing or something

wardi: it's dead code, it does nothing but it looks like it might do something, likely leading to user confusion. Why would you keep it? Should probably be getting flagged by our linter too

dborzov: OK, true that :)

wardi:
@dborzov go ahead and merge after removing that dead code and fixing the merge conflict

@JackMc do you know why our linter isn't picking up unused variables?

JackMc:
Hm... I'll look into it!

JackMc:
Pyflakes has been kinda quiet lately..

Issue #10, opened 2015-01-03
Add -v options to see output of common commands


deniszgonjanin:
Like `datacats install`, `datacats create`, and `datacats deploy`

dborzov:
What is it supposed to show specifically?

I think, at the moment it channels the `stderr` of remote command being executed to datacat command's `stdout`. Is that a temporary thing or the longer term design decision? Should it channel the `stderr` only with the `verbose` key on? 

@wardi , @JackMc any thoughts?

JackMc:
A lot of the things that are run within those commands are ran with their stdout going to the docker logs (accessible through "datacats logs"). I think what is being asked here is that there be a way (a -v --verbose option) to see this output directly on the console.

wardi:
datacats deploy is now verbose (it shows the rsync output) it could be used as an example of a verbose command. I think install should be verbose by default too, but not during create - that one *should* always work without downloading anything.

So I'd say, let's not bother with a --verbose let's just make install verbose by default too

JackMc:
So the install step of create shouldn't be verbose, but an explicit command-line use of install should be verbose?

wardi:
Yeah, that makes sense to me

JackMc:
:+1:

wardi:
we have better deploy errors #157 and verbose installs by default now #158

PR #158, opened 2015-06-04
Verbose installs

JackMc:
Addresses #10

### DIFF
--- a/datacats/cli/install.py
+++ b/datacats/cli/install.py
@@ -7,6 +7,7 @@
 import sys
 from os import listdir
 from os.path import isdir, exists
+from datacats.docker import container_logs
 
 from datacats.cli import manage
 from datacats.docker import check_connectivity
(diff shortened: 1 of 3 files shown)


datacats/cli/install.py
```diff
@@ -71,9 +73,13 @@ def install_all(environment, clean):
 
     for s in ['ckan'] + sorted(srcdirs):
         write('Installing ' + s)
-        environment.install_package_develop(s)
+        c_id = environment.install_package_develop(s)
+        if verbose and c_id:
+            print '\n'.join(container_logs(c_id, "all", True, None))
```
wardi: we should be using the generator to get the terminal output as it arrives, not all at once after completion. Some of these installs are really slow.

JackMc: :+1:

JackMc:
Ready for CR

JackMc:
Doesn't do anything - duly noted!

Issue #159, opened 2015-06-05
Linter doesn't see unused parameters


JackMc:

JackMc:
Whatever magical technology pycharm uses does this

JackMc:
Trying to find recommendations

PR #160, opened 2015-06-05
Replace pyflakes with pylint as our static analysis tool

JackMc:
Addresses #159


JackMc:
Some of dat sweet sweet CR :)

wardi:
@JackMc please go ahead and merge after resolving the conflict

PR #165, opened 2015-06-09
Add datapusher support

JackMc:
Addresses #9


JackMc:
To do:
- [ ] Change site_url in the case where we're not on prod and are on Linux
- [ ] Figure out what redis has to do with datapusher

deniszgonjanin:
This runs datapusher on a separate container but uses the web image?

By default datapusher uses a sqlite db for jobs. Perhaps this can be switched to use postgres in the existing postgres container. You don't need redis

Also, I see you are using default datapusher settings, which runs it on flask dev server and does not allow concurrent connections. That would need addressing. Tornado is an option

JackMc:
Yes, we discussed and thought a separate image would be too big.

Okie dokie.

It's using your fork, what magic do I do to make it run with tornado?

deniszgonjanin:
cool, reusing the web container is a good idea

I forget what needs to happen with the tornado fork - I think it maybe just works<sup>:tm:</sup>

wardi:
yeah, redis was my fault. That's for harvester not datapusher. Let's remove it.

I'm still curious how ckan-service-provider doesn't have a process sitting there polling if all it has is a database.

We certainly could use the same postgres database if that's supported. Then we'd at least have a chance of resuming if an environment is stopped before the datapusher job is complete

JackMc:
I can't get CI to consistently pass or fail

JackMc:
I think there's some crazy race condition

JackMc:
Looks like it's failing because the venv isn't mounted:

```
/scripts/datapusher.sh: line 3: /usr/lib/ckan/bin/activate: No such file or directory
  File "datapusher/main.py", line 3, in <module>
    import ckanserviceprovider.web as web
ImportError: No module named ckanserviceprovider.web
lxc-start: The container failed to start.
lxc-start: Additional information can be obtained by setting the --logfile and --logpriority options.
```

But this only happens sometimes. Any ideas?

JackMc:
Right, that should fix that.

JackMc:
Yay it works! :raised_hands:

JackMc:
Merged in testing stuffs :smile:

JackMc:
To do:
- [ ] Fix merge conflict
- [ ] site_url needs to be patched

deniszgonjanin:
I get the following error on create:
```
$ datacats create datapusher
Creating environment "datapusher/primary"....
  File "/Users/dz/Envs/datacats/bin/datacats", line 9, in <module>
  File "/Users/dz/source/datacats/datacats/cli/main.py", line 84, in main
    return command_fn(opts)
  File "/Users/dz/source/datacats/datacats/cli/create.py", line 52, in create
    datapusher=not opts['--no-datapusher'],
  File "/Users/dz/source/datacats/datacats/cli/create.py", line 91, in create_environment
    fn()
  File "/Users/dz/source/datacats/datacats/cli/create.py", line 74, in <lambda>
    lambda: environment.create_source(datapusher),
  File "/Users/dz/source/datacats/datacats/environment.py", line 499, in create_source
    image=self._preload_image())
  File "/Users/dz/source/datacats/datacats/docker.py", line 180, in web_command
    raise WebCommandError(command, c['Id'][:12], logs)
TypeError: __init__() takes exactly 3 arguments (4 given)
```

JackMc:
Huh. Ian was getting a similar problem on one of his PRs. I'll look into it.

deniszgonjanin:
On an existing environment, I run into the following:
```
$ datacats start
  File "/Users/dz/Envs/datacats/bin/datacats", line 9, in <module>
  File "/Users/dz/source/datacats/datacats/cli/main.py", line 91, in main
    return command_fn(environment, opts)
  File "/Users/dz/source/datacats/datacats/cli/manage.py", line 62, in start
    reload_(environment, opts)
  File "/Users/dz/source/datacats/datacats/cli/manage.py", line 98, in reload_
    log_syslog=opts['--syslog'])
  File "/Users/dz/source/datacats/datacats/environment.py", line 700, in start_ckan
    datapusher=datapusher)
  File "/Users/dz/source/datacats/datacats/environment.py", line 785, in _run_web_container
    log_syslog=log_syslog
  File "/Users/dz/source/datacats/datacats/docker.py", line 266, in run_container
    port_bindings=port_bindings)
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/docker_py-1.2.2-py2.7.egg/docker/utils/decorators.py", line 15, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/docker_py-1.2.2-py2.7.egg/docker/client.py", line 1063, in start
    self._raise_for_status(res)
  File "/Users/dz/Envs/datacats/lib/python2.7/site-packages/docker_py-1.2.2-py2.7.egg/docker/client.py", line 122, in _raise_for_status
    raise errors.APIError(e, response, explanation=explanation)
docker.errors.APIError: 500 Server Error: Internal Server Error ("Cannot start container a5cfa9e5b3ccbef156667df2d6d6af04ec23f6ce8176a1177a004368e16ad6a5: Cannot link to a non running container: /datacats_datapusher_myckan_primary AS /datacats_web_myckan_primary/datapusher")
```

deniszgonjanin:
@JackMc thanks, very excited about this

JackMc:
Seems like this is an issue with the Pylint cleanup

Time to bring back the container_id parameter :smile:

JackMc:
Yeah, merging with master seems to have done something to the datapusher folder. Hm.

JackMc:
@deniszgonjanin Re-run "datacats pull", your Docker images are out of date and don't contain the datapusher source in the preload image.

We should get some kinda versioning thing going on.

deniszgonjanin:
@JackMc yea, we should probably be tagging those docker images more frequently.

deniszgonjanin:
Creating a new environment starts a datapusher container now, but the container cannot connect back to the web container to post the data:
```
Job "push_to_datastore (trigger: RunTriggerNow, run = True, next run at: None)" raised an exception
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/apscheduler/scheduler.py", line 512, in _run_job
    retval = job.func(*job.args, **job.kwargs)
  File "/project/datapusher/datapusher/jobs.py", line 222, in push_to_datastore
    resource = get_resource(resource_id, ckan_url, api_key)
  File "/project/datapusher/datapusher/jobs.py", line 178, in get_resource
    'Authorization': api_key}
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/requests/api.py", line 88, in post
    return request('post', url, data=data, **kwargs)
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/requests/api.py", line 44, in request
    return session.request(method=method, url=url, **kwargs)
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/requests/sessions.py", line 456, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/requests/sessions.py", line 559, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/requests/adapters.py", line 375, in send
    raise ConnectionError(e, request=request)
ConnectionError: HTTPConnectionPool(host='boot2docker', port=5830): Max retries exceeded with url: /api/3/action/resource_show (Caused by <class 'socket.gaierror'>: [Errno -2] Name or service not known)
```

To test this:
- Create a new datacats environment
- Create a new dataset and upload a .csv resource
- Go to /dataset/<name>/resource_data/<resource_id> (see below image for details)

screen shot 2015-06-11 at 3 23 39 pm

JackMc:
Darn it... :\ This worked on my machine...

wardi:
@deniszgonjanin did you set a site_url with `boot2docker` in it?

deniszgonjanin:
I think I may know what the issue is. I have in my /etc/hosts: `192.168.59.103 boot2docker`

So all my datacats envs are running on http://boot2docker. Containers don't know about this domain

wardi:
And your boot2docker env vars use that name for your boot2docker vm.  I guess we'll have to resolve the host name and use that IP address in the site_urls we generate for boot2docker.

deniszgonjanin:
Or we could do a lookup for the Docker IP and see if there is a domain name set, then set that domain name in the datapusher container?

wardi:
I don't think you get reverse mapping (IP->name) when you add things to your hosts file. Another option might be mounting the user's /etc/hosts into the datapusher container, but I don't know if that conflicts with any docker magic

wardi:
Or is this our cue to finally investigate docker machine as a replacement for boot2docker?

JackMc:
Does docker machine handle this case?

wardi:
@JackMc no idea, but we've spent plenty of time on boot2docker issues, and when we ask it seems that the new recommendation is to use docker machine instead. No idea how docker machine handles volume mounts though (it's supposed to work just as well remote as locally)

deniszgonjanin:
As an aside, if I change my site_url in the ini to `http://192.168.59.103:5830/`, I still get the same error:
```
ConnectionError: HTTPConnectionPool(host='boot2docker', port=5830): Max retries exceeded with url: /api/3/action/resource_show (Caused by <class 'socket.gaierror'>: [Errno -2] Name or service not known)
```

By what secret magic is this happening?

wardi:
@deniszgonjanin you need to set site_url in .datacats-environment to override the one set by datacats. Magic is here: https://github.com/datacats/datacats/blob/master/datacats/environment.py#L692

deniszgonjanin:
@wardi cool, that works. Maybe we can just override the site_url for now, unless that has unintended consequences

wardi:
@deniszgonjanin I'm not sure what you mean. datacats always overrides the site_url. Do you think we don't need to make a special case for boot2docker users that are using a hostname they've added to their /etc/hosts?

deniszgonjanin:
No, I mean let's override it for them. Would that present any problems?

wardi:
resolving the host name to an IP and using that for the site_url in the boot2docker case would be safe. Inserting the host name into the container would be slightly prettier but seems like a lot of work and could possibly cause problems

deniszgonjanin:
I agree. Let's do what you suggest then, until we can investigate docker-machine

JackMc:
So would an appropriate algo be:
 1. Create site
 2. Set site_url to (boot2docker_resolved_ip if is_boot2docker() else container_ip)
 3. server can override site_url in INI if necessary
?

JackMc:
The new commit should implement the above

wardi:
@JackMc please merge master into this branch so we can see if the tests are passing

JackMc:
@wardi Merged in, tests pass.

deniszgonjanin:
let's bring this up to date with master when there's a chance

JackMc:
Already working on it :)

JackMc:
Please test - I think we should merge this soon! It works for me.

deniszgonjanin:
Great, testing now

deniszgonjanin:
Merge it

makkus:
I'm using Linux (so there is no need for boot2docker), and I'm getting the 'Error: Process completed but unable to post to result_url' message. Is there a workaround for my situation?
Tried to change the [site_primary]/address field in .datacats-environment to a hostname that is in /etc/hosts on the host, but that didn't help. Any suggestions?

JackMc:
Hiya @makkus, thanks for the report. Can you show me the output of the following commands (execute while in the source directory)?

``datacats shell cat /project/development.ini``
``docker ps -a``

JackMc:
This is related to a problem where the data pusher container will get confused and try to contact itself on the Linux development case - as the site_url is set to 127.0.0.1

wardi:
@makkus also on linux you need to have ICC enabled in your docker configuration and you might need to check that your firewall rules (if you have any) aren't blocking connections between docker containers.

makkus:
Yes, I had the same problem when I was using ckan-docker, had to manually change the host file on the datapusher container because of the no-circular-links restriction in docker/docker-compose.

With a vanilla environment (127.0.0.1 in .datacats-environment) I get:

-------------------------------------------------------------------------------
$ datacats shell . cat /project/development.ini
[DEFAULT]
debug = true

[server:main]
use = egg:Paste#http
host = 0.0.0.0
port = 5000

[app:main]
use = egg:ckan
full_stack = true
cache_dir = /tmp/%(ckan.site_id)s/
beaker.session.key = ckan
beaker.session.secret = QM2fSQtinqOeUV7k
app_instance_uuid = {0094a8dc-4cf9-4256-b148-7b578990545d}
who.config_file = %(here)s/who.ini
who.log_level = warning
who.log_file = %(cache_dir)s/who_log.ini
sqlalchemy.url = postgresql://ckan:KWkHWMMH3uQbVnBZ@db:5432/ckan
ckan.datastore.write_url = postgresql://ckan_datastore_readwrite:nByaYzftzz2xhXSD@db:5432/ckan_datastore
ckan.datastore.read_url = postgresql://ckan_datastore_readonly:kR5jDCp4LiUwACom@db:5432/ckan_datastore
ckan.datastore.default_fts_lang = english
ckan.datastore.default_fts_index_method = gist
ckan.site_url = http://127.0.0.1:5439
ckan.auth.anon_create_dataset = false
ckan.auth.create_unowned_dataset = true
ckan.auth.create_dataset_if_not_in_organization = true
ckan.auth.user_create_groups = true
ckan.auth.user_create_organizations = true
ckan.auth.user_delete_groups = true
ckan.auth.user_delete_organizations = true
ckan.auth.create_user_via_api = false
ckan.auth.create_user_via_web = false
ckan.auth.roles_that_cascade_to_sub_groups = admin
ckan.site_id = default
solr_url = http://solr:8080/solr
ckan.plugins = datastore resource_proxy text_view datapusher recline_grid_view recline_graph_view catstown_theme pages
ckan.views.default_views = image_view text_view recline_view
ckan.site_title = catstown
ckan.site_logo = 
ckan.site_description = 
ckan.favicon = /images/icons/ckan.ico
ckan.gravatar_default = identicon
ckan.preview.direct = png jpg gif
ckan.preview.loadable = html htm rdf+xml owl+xml xml n3 n-triples turtle plain atom csv tsv rss txt json
ckan.locale_default = en
ckan.locale_order = en pt_BR ja it cs_CZ ca es fr el sv sr sr@latin no sk fi ru de pl nl bg ko_KR hu sa sl lv
ckan.locales_offered = 
ckan.locales_filtered_out = en_GB
ckan.feeds.authority_name = 
ckan.feeds.date = 
ckan.feeds.author_name = 
ckan.feeds.author_link = 
ckan.storage_path = /var/www/storage
ckan.datapusher.url = http://datapusher:8800
ckan.hide_activity_from_users = %(ckan.site_id)s

[loggers]
keys = root, ckan, ckanext

[handlers]
keys = console

[formatters]
keys = generic

[logger_root]
level = WARNING
handlers = console

[logger_ckan]
level = INFO
handlers = console
qualname = ckan
propagate = 0

[logger_ckanext]
level = DEBUG
handlers = console
qualname = ckanext
propagate = 0

[handler_console]
class = StreamHandler
args = (sys.stderr,)
level = NOTSET
formatter = generic

[formatter_generic]
format = %(asctime)s %(levelname)-5.5s [%(name)s] %(message)s
---------------------------------------------------------------------------------------
Also:

$  docker ps   
CONTAINER ID        IMAGE                      COMMAND                CREATED             STATUS              PORTS                      NAMES
9726f18698f3        datacats/web:latest        "/scripts/web.sh Fal   About an hour ago   Up About an hour    127.0.0.1:5439->5000/tcp   datacats_web_catstown_primary          
ea16ad13f8fa        datacats/web:latest        "/scripts/datapusher   About an hour ago   Up About an hour    5000/tcp                   datacats_datapusher_catstown_primary   
0009a8ec3900        datacats/solr:latest       "/usr/share/tomcat6/   About an hour ago   Up About an hour    8080/tcp                   datacats_solr_catstown_primary         
d30a371e5cd9        datacats/postgres:latest   "/docker-entrypoint.   About an hour ago   Up About an hour    5432/tcp                   datacats_postgres_catstown_primary

JackMc:
Thank you! We try to patch the site_url so that it points at the web container's API on the docker0 interface, but that doesn't seem to be working.

makkus:
Default ICC behaviour is 'true', right? So that should be fine sine I didn't change that. Also, firewall should be fine too, since all of my other projects' docker-compose links works.

JackMc:
I will look into this tomorrow morning, as I left my charger at work and my laptop's dying :cry:

makkus:
Happy to do some more debugging. If you point me to where you do the site_url rewrite I can look a bit closer myself.
Also, no rush. Datapusher is not critical for the part of the setup I'm working at the moment.

JackMc:
@makkus That'd be great! The script that is **supposed** to do the patching is called adjust_devini.py in the datacats/scripts directory.

JackMc:
The possible problems I can think of are the following:
* Script is failing silently for some reason
* Script isn't being mounted on the volume and therefore isn't getting executed
* The web.sh script is not being executed with it's second parameter True (indicating a Linux development environment) https://github.com/datacats/datacats/blob/master/datacats/scripts/web.sh#L22 is where this param is used and https://github.com/datacats/datacats/blob/master/datacats/environment.py#L385 is where it is being decided if we should pass in a True or False for that particular parameter

wardi:
@makkus did you put `site_url=http://127.0.0.1...' in your .datacats-environment file? That certainly won't work because the datapusher container won't be able to reach the ckan server with that address. either remove your site_url setting from .datacats-environment or use an address that can be reached from a docker container running locally (like an IP on one of your other network interfaces)

makkus:
Looking at the adjust_devini.py script within the container:

    #!/usr/bin/env python2

    from ConfigParser import SafeConfigParser
    import sys

    cp = SafeConfigParser()
    cp.read('/project/development.ini')

    cp.set('app:main', 'ckan.site_url', 'http://{}:{}'.format(sys.argv[1], sys.argv[2]))

    with open('development.ini', 'w') as fp:
        cp.write(fp)

When opening the file to write, it doesn't use an absolute path, could that be an issue? Not sure where the script is called from, so it might or might not work depending on that...

makkus:
Aplogies, the output to the 'cat' command above is misleading, probably due to me playing around with the .datacats-environment file. After I deleted the whole folder, and recreated a new one I'm seeing this value for the ckan.site_url key:

    ckan.site_url = http://None:5859

makkus:
@wardi No, I didn't put site_url in the .datacats-environment field, I only changed the value of the

    [site_primary]
    address = xxx

field. I tried the address of the docker0 bridge, as well as the ip address of my ethernet adapter, both to no avail.

JackMc:
Can you check if this patch fixes it? It does on my Linux VM.

```
From cfefb2321b4e4b9e24ed0fb3580c86b9e7038a42 Mon Sep 17 00:00:00 2001
From: Jack McCracken <jackamc19@gmail.com>
Date: Fri, 26 Jun 2015 16:17:44 -0400
Subject: [PATCH] Fix adjust_devini for right port and absolute development.ini
 path

---
 datacats/scripts/adjust_devini.py | 2 +-
 datacats/scripts/web.sh           | 2 +-
 2 files changed, 2 insertions(+), 2 deletions(-)

diff --git a/datacats/scripts/adjust_devini.py b/datacats/scripts/adjust_devini.py
index ab2f9f1..d1f7afa 100755
--- a/datacats/scripts/adjust_devini.py
+++ b/datacats/scripts/adjust_devini.py
@@ -8,5 +8,5 @@ cp.read('/project/development.ini')
 
 cp.set('app:main', 'ckan.site_url', 'http://{}:{}'.format(sys.argv[1], sys.argv[2]))
 
-with open('development.ini', 'w') as fp:
+with open('/project/development.ini', 'w') as fp:
     cp.write(fp)
diff --git a/datacats/scripts/web.sh b/datacats/scripts/web.sh
index 1c86883..7e51282 100755
--- a/datacats/scripts/web.sh
+++ b/datacats/scripts/web.sh
@@ -20,7 +20,7 @@ fi
 while true; do
 	# fix our development.ini
 	if [ "$2" = "True" ]; then
-		/scripts/adjust_devini.py "$(ip route get 8.8.8.8 | awk 'NR==1 {print $NF}')" "$port"
+		/scripts/adjust_devini.py "$(ip route get 8.8.8.8 | awk 'NR==1 {print $NF}')" "5000"
 	fi
 
 	# production
-- 
1.9.1
```

wardi:
@JackMc pls add a regression test when you write the PR

makkus:
Ok, I did apply the patch, but still got the ckan.site_url = http://None:[port] value. Funnily enough, after a 'datacats reload' the value is http://127.0.0.1:[port]

Inserted some debug statements, and I can confirm that both the production as well as self.always_prod have "False" as value (in environment.py). Haven't dug any deeper yet, might do more debugging if you guys can't see straight away what's wrong, seeing you know the codebase much better.

Also, looking at https://github.com/datacats/datacats/blob/master/datacats/scripts/web.sh:

When doing the bash tests for string equality, the first time around it is done using 'true' (lowercase), then 'True' is used. Also, the last test uses '==' instead of '='. Don't think those are related to my issue, but may break other stuff.

makkus:
Sorry, did get confused with the order of the arguments for the web.sh script. So, I now think the issue is that in environment.py the self.address variable is set to None, but is only tested against the string '127.0.0.1' ( https://github.com/datacats/datacats/blob/master/datacats/environment.py#L386 ). 

It kinda feels that this variable probably should be set at some stage and isn't, but I guess you guys will know. Either that, or the check needs to be expanded to include a test against 'None'.

JackMc:
Hmmm. The site_url should get mounted over in the container with a slightly modified one. The actual one is in ``~/.datacats/myenv/sites/primary/run/development.ini``. Can you check that file?

JackMc:
It's really interesting that it's not being set for you. It should always be set to at least some value...

JackMc:
Got it, I think. The address wasn't being passed into the new function at https://github.com/datacats/datacats/blob/master/datacats/cli/create.py#L60

JackMc:
```
From 56e0589ee2faf7e1db21efa4a713e105c54d8519 Mon Sep 17 00:00:00 2001
From: Jack McCracken <jackamc19@gmail.com>
Date: Mon, 6 Jul 2015 10:27:51 -0400
Subject: [PATCH] Fixes as discussed in PR about datapusher

---
 datacats/cli/create.py            | 2 +-
 datacats/scripts/adjust_devini.py | 2 +-
 datacats/scripts/web.sh           | 2 +-
 3 files changed, 3 insertions(+), 3 deletions(-)

diff --git a/datacats/cli/create.py b/datacats/cli/create.py
index 1d03cf8..9646e86 100644
--- a/datacats/cli/create.py
+++ b/datacats/cli/create.py
@@ -57,7 +57,7 @@ def create_environment(environment_dir, port, ckan_version, create_skin, site_na
         start_web, create_sysadmin, address, log_syslog=False, datapusher=True, quiet=False):
     # pylint: disable=unused-argument
     # FIXME: only 2.3 preload supported at the moment
-    environment = Environment.new(environment_dir, '2.3', site_name, port=port)
+    environment = Environment.new(environment_dir, '2.3', site_name, address=address, port=port)
 
     try:
         # There are a lot of steps we can/must skip if we're making a sub-site only
diff --git a/datacats/scripts/adjust_devini.py b/datacats/scripts/adjust_devini.py
index ab2f9f1..d1f7afa 100755
--- a/datacats/scripts/adjust_devini.py
+++ b/datacats/scripts/adjust_devini.py
@@ -8,5 +8,5 @@ cp.read('/project/development.ini')
 
 cp.set('app:main', 'ckan.site_url', 'http://{}:{}'.format(sys.argv[1], sys.argv[2]))
 
-with open('development.ini', 'w') as fp:
+with open('/project/development.ini', 'w') as fp:
     cp.write(fp)
diff --git a/datacats/scripts/web.sh b/datacats/scripts/web.sh
index 1c86883..7e51282 100755
--- a/datacats/scripts/web.sh
+++ b/datacats/scripts/web.sh
@@ -20,7 +20,7 @@ fi
 while true; do
 	# fix our development.ini
 	if [ "$2" = "True" ]; then
-		/scripts/adjust_devini.py "$(ip route get 8.8.8.8 | awk 'NR==1 {print $NF}')" "$port"
+		/scripts/adjust_devini.py "$(ip route get 8.8.8.8 | awk 'NR==1 {print $NF}')" "5000"
 	fi
 
 	# production
-- 
2.4.5
```

Please apply to a clean upstream/master

makkus:
Yup. Can confirm that works. Both before and after a datacats reload.

JackMc:
Thank you. Sorry for the inconvenience and if you have any other problems report an issue and hopefully we can help out! :)

PR #169, opened 2015-06-10
CLI help text fixes

wardi:


JackMc:
Didn't know you could do that! (Extend a description onto a new line)

JackMc:
:+1:

datacats/cli/create.py
```diff
@@ -97,14 +96,14 @@ def init(opts):
     """Initialize a purged environment or copied environment directory
 
 Usage:
-  datacats init [-in] [--syslog] [-s NAME] [--address=IP] [ENVIRONMENT_DIR [PORT]]
-  datacats init [-ni] [--syslog] [--address=IP] [ENVIRONMENT_DIR [PORT]]
+  datacats init [-in] [--syslog] [-s NAME] [--address=IP]
+                [ENVIRONMENT_DIR [PORT]]
```
dborzov: Hey, I know it is bikeshedding and not that important :smiley:,

but this command's name `init` sounds somewhat confusing too me taking into account what it does. 

Within the context of CLI tools, one would assume that `init` does either of these two things:
1) setup computer for work with the tool (e.g. pull the needed docker images and set up `~/.datacats/user_profile`)
2) do what `datacats create` does

What do you guys think?

JackMc: Hm, I can see the confusion.

Any suggestions for a better name?

wardi: I think `init` is a good fit for what it does. It's like `def __init__` in python where we have a class and now we're creating an instance from it. It also sounds like less work than `create` (fewer letters :smile:)

JackMc: How about "creat" :stuck_out_tongue:

wardi: That's brilliant! 

... No. :smirk:

PR #174, opened 2015-06-10
Catch and wrap with user-friendly message all unspecified Errors

dborzov:
When unspecified (programming) error is encountered, we output an error message like this that prompts the user to report it here as an issue:

Img

Before it was just spilling out the traceback as it is.

### DIFF
--- a/datacats/error.py
+++ b/datacats/error.py
@@ -11,11 +11,11 @@ class DatacatsError(Exception):
                 "type_description": parent_exception.user_description,
                 "message": parent_exception.__str__(),
             }
-            self.message = "".join(["{original}\n\n",
+            self.message = "".join([colored.blue("{original}\n\n").__str__(),
                                     "~" * 30,
                                     "\n{type_description}:\n",
-                                    "{message}",
-                                    "~" * 30, "\n"]).format(**vals)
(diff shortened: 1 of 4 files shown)


JackMc:
Can we merge this? :smile:

datacats/error.py
```diff
@@ -11,11 +11,11 @@ def __init__(self, message, format_args=(), parent_exception=None):
                 "type_description": parent_exception.user_description,
                 "message": parent_exception.__str__(),
             }
-            self.message = "".join(["{original}\n\n",
+            self.message = "".join([colored.blue("{original}\n\n").__str__(),
```
JackMc: Why not str(colored.blue(blah))?

dborzov: Good point! Tnx!

dborzov:
Yeah, let's merge it! 
It includes turning off this annoying SSL warnings (when communicating with docker) snd that thing really needs to be on master.

JackMc:
:+1:

datacats/error.py
wardi: please merge after str() changes above and below

wardi:
turning off the extra SSL warnings might be nice to backport to the current stable release too

Issue #171, opened 2015-06-10
Migration warning should be explained better


deniszgonjanin:
This message should be expanded on. It tells me that important changes are happening, but doesn't tell me why I should care, or in which cases I may be unwise to perform the migration. It gives the user a choice to make, without explaining any implications of that choice.

```
This migration will change the format of your datadir. Are you sure? (y/n) [n]:
```

JackMc:
The only problem with being more specific is that we allow for migrations from 1->2 and 2->1 and eventually n_1->n_2

JackMc:
So anything more specific may be incorrect in the 2->1 direction

deniszgonjanin:
Maybe just explaining why migrating would ever be dangerous or potentially destructive. As a user, is there a scenario where I should think twice before migrating?

JackMc:
#176 fixes this

PR #176, opened 2015-06-11
Better error handling for migrations

JackMc:
Addresses #171


wardi:
instead of a generic message why not show the docstring of the migration function that's being called? Then you can be specific e.g.: "After migration this environment will work with datacats version >= 1.0.0" or "After migration only the primary site will be kept and this environment will work with datacats versions < 1.0.0"

JackMc:
Good idea! :smile:

Issue #162, opened 2015-06-05
Spatial Extension Support


JackMc:
This will require us to run the postgis scripts on our postgres container.

PR #177, opened 2015-06-11
Add postgis to the postgres container, allowing for spatial extension support

JackMc:
Addresses #162


JackMc:
Looks like the LXC driver doesn't support exec :cry:

wardi:
ugh, well we could stop the postgres container, run that one command in a new container then start it again.

deniszgonjanin:
In general it's better to do things that way. Don't use exec unless necessary

JackMc:
Alternatively we could install psql client on the web container and use a web_command

wardi:
@JackMc that seems less pleasant: we would need the postgis sql stuff in the web container too.

JackMc:
Well then we could take the postgis script stuff out of the postgres container

wardi:
@JackMc no, I think that's required too (some of it references tools installed on the postgres site) I'm pretty sure stopping the container, running the one command and restarting is the simplest thing that will also work on circleci

JackMc:
Okie dokie

deniszgonjanin:
If I rebuild the postgres docker image, should this work with existing environments? Do I need to run that docker command manually?

The following check from the [docs](http://docs.ckan.org/projects/ckanext-spatial/en/latest/install.html#install-the-extension) fails after installing and enabling ckanext-spatial:
```
psql -d ckan -c "SELECT postgis_full_version()"
ERROR:  function postgis_full_version() does not exist
LINE 1: SELECT postgis_full_version()
               ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

JackMc:
It won't work with previously existing DataCats instances, as it doesn't run the ``/scripts/install_postgis.sh`` script. You can mount and run this manually if you'd like (for now) or we could add a command for this to DataCats

deniszgonjanin:
That's ok, running manually is fine for my immediate purposes. I'm doing some tests but this looks good so far. We should have a migration for when this gets merged though

JackMc:
M'kay, I didn't think this was worth upping the format version

deniszgonjanin:
Talk to @wardi , I'm ok either way. If you don't up the version, put instructions in the FAQ about how to upgrade manually

JackMc:
Okie dokie :)

deniszgonjanin:
If I do:
```
$ datacats stop
$ datacats paster db clean
$ datacats paster db init
$ datacats reload
```

I get the error:
```
  File "/usr/lib/ckan/bin/paster", line 9, in <module>
    load_entry_point('PasteScript==1.7.5', 'console_scripts', 'paster')()
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/paste/script/command.py", line 104, in run
    invoke(command, command_name, options, args[1:])
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/paste/script/command.py", line 143, in invoke
    exit_code = runner.run(args)
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/paste/script/command.py", line 238, in run
    result = self.command()
  File "/project/ckan/ckan/lib/cli.py", line 683, in command
    self._load_config()
  File "/project/ckan/ckan/lib/cli.py", line 148, in _load_config
    load_environment(conf.global_conf, conf.local_conf)
  File "/project/ckan/ckan/config/environment.py", line 232, in load_environment
    p.load_all(config)
  File "/project/ckan/ckan/plugins/core.py", line 134, in load_all
    load(*plugins)
  File "/project/ckan/ckan/plugins/core.py", line 166, in load
    plugins_update()
  File "/project/ckan/ckan/plugins/core.py", line 116, in plugins_update
    environment.update_config()
  File "/project/ckan/ckan/config/environment.py", line 361, in update_config
    plugin.configure(config)
  File "/project/ckanext-spatial/ckanext/spatial/plugin.py", line 50, in configure
    setup_model()
  File "/project/ckanext-spatial/ckanext/spatial/model/package_extent.py", line 32, in setup
    'Please refer to the "Setting up PostGIS" section in the README.')
Exception: The spatial extension is enabled, but PostGIS has not been set up in the database. Please refer to the "Setting up PostGIS" section in the README.
```

So the setup script would need to be ran every time. Perhaps having a command to do it is the easiest? How can we smooth out the process? cc/ @wardi

JackMc:
We can't trap db inits or db cleans (well we could but that'd be super hacky) so if that's to work it might be better to advise users to use the ``datacats reset`` option in #178 to reset their database on their site

deniszgonjanin:
Ok, I like that. Only thing that sucks is that the reset command takes quite a long time when you have a bunch of extensions installed. But whatever, I think that's good enough. Let me test how that workflow feels

JackMc:
Reset shouldn't do an install, it simple purges and reinitializes the DB

deniszgonjanin:
Oh ic. :+1:

datacats/docker.py
```diff
@@ -138,6 +138,23 @@ def binds_to_volumes(volumes):
...
+    :param stderr: True to hook up stderr to the container's stderr
+    :param tty: True to allocate a tty
+    :param detach: True to detach from the current process
+    :param stream: True to return a generator of response data
+    :return:
+    """
+    c = _get_docker().exec_create(container_name, command, stdout, stderr, tty)
+    _get_docker().exec_start(c['Id'], detach, tty, stream)
+
+
```
wardi: this is dead code now?

wardi:
merge now, fix dead code later

deniszgonjanin:
:scream:

datacats/environment.py
```diff
@@ -608,14 +615,27 @@ def ckan_db_init(self, retry_seconds=DB_INIT_RETRY_SECONDS):
...
                     clean_up=True,
                     )
+                self.stop_postgres_and_solr()
+                container = run_container(
+                    name=self._get_container_name('postgres'),
+                    image='datacats/postgres',
+                    environment=self.passwords,
+                    ro={INSTALL_POSTGIS: '/scripts/install_postgis.sh'},
+                    rw=rw,
+                    volumes_from=volumes_from)
```
wardi: oops. We're not actually running the command here are we?

JackMc: Oh. My bad!

JackMc:
Dead code was intentional b.c. it's useful :)

datacats/environment.py
JackMc: Pushed a fix to master

Issue #170, opened 2015-06-10
Idea: datacats reset command


wardi:
Delete the files, reset the db and recreate the admin user.  sort of like purge + init but just for one site

JackMc:
So basically equivalent to:
```
datacats purge -s mysite myenv
datacats init -s mysite myenv
```

??

wardi:
yes, except without the install step and if it's the last site it won't destroy the venv and reinstall everything

JackMc:
Ahh ok makes sense :)

wardi:
I don;t know if this makes more sense as a separate command or an option on `init`

dborzov:
Let's drop the current `init` completely, and have `reset` with options to replicate what `init` does now.

`init` as a CLI command is now an idiom that stands for completely different things. Let's count all the examples:
- `git init`
-`heroku init`
- `boot2docker init`
- `ipfs init`

and the list goes and goes.

`init` is expected to either:
1) setup the computer for work with the tool (e.g. pull the needed docker images and set up ~/.datacats/user_profile in our case)
2) or do what `datacats create` does, that is create an instance or an environment

JackMc:
That might be a good idea.

wardi:
`reset` doesn't make sense as the way to add new sites to an environment

wardi:
`datacats assemble!`

![thundercats assemble](http://i54.tinypic.com/2mfcawk.gif)

dborzov:
:joy::joy::joy: `datacats assemble!` is the only command we will ever need

JackMc:
Does the CKAN ``paster --plugin=ckan db init`` command reset the database (i.e. get rid of everything)?

deniszgonjanin:
No, `paster --plugin=ckan db clean` does that

JackMc:
Thank you! :)

wardi:
Might need a `paster --plugin=ckan search-index clear` too (I think these commands should be merged)

wardi:
If we're implementing a reset I wouldn't trust any of these things. I'd just nuke the directories and create them again.

PR #178, opened 2015-06-11
Add a 'reset' command to datacats

JackMc:
This command is as described in #170

### DIFF
--- a/datacats/cli/create.py
+++ b/datacats/cli/create.py
@@ -104,7 +104,7 @@ Options:
     # pylint: disable=unused-argument
     print 'Re-initializing database...'
     environment.start_postgres_and_solr()
-    environment.ckan_db_init()
+    environment.paster_command("db clean")
     adminpw = confirm_password()
     environment.create_admin_set_password(adminpw)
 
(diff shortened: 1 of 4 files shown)


JackMc:
Should we make the -s option mandatory (as it is now) or optional, defaulting to 'primary'?

JackMc:
Any opinions on this?

wardi:
optional, defaulting to 'primary' like the other commands

JackMc:
Done

datacats/cli/create.py
```diff
@@ -92,7 +92,32 @@ def create_environment(environment_dir, port, ckan_version, create_skin, site_na
         raise
 
 
-def init(opts):
+def reset(environment, opts):
+    """Resets a site to the default state. This will re-initialize the
+database and recreate the administrator account.
+
+Usage:
+  datacats reset [-s NAME] ENVIRONMENT
```
wardi: ENVIRONMENT should be optional like the other commands

wardi:
merge now fix ENVIRONMENT command later

PR #188, opened 2015-06-15
Update Install instructions on README.md

dborzov:
Oh, and for the Installation section on README.md it probably should:
- point people who want to just use datacats at the doc page
- point people who want to hack datacats to `python setup.py develop`

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,14 +9,15 @@ any platform, along with a command to deploy that exact environment to the cloud
 
 
 ## Install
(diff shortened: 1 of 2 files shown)


dborzov:
This probably should have been in the previous PR and does not warrant an extra 'merge' commit. Let me know if its OK, and will fast-forward merge it on master.

wardi:
"ruse" a custom branch?

Also, this looks like it should be in with the main install instructions because it's tempting to just copy+paste the commands under the install heading without reading the paragraphs above

dborzov:
@wardi fixed

Issue #190, opened 2015-06-15
exceptions in exceptions hiding real traceback?


wardi:
```
$ datacats reset .
Resetting...
Creating environment bare/primary from existing environment directory "bare".....
Initializing database----------------------------------------
datacats: problem was encountered:
Traceback (most recent call last):
  File "/home/ian/boxkite/venv/bin/datacats", line 9, in <module>
    load_entry_point('datacats==1.0.0dev1', 'console_scripts', 'datacats')()
  File "/home/ian/boxkite/git/datacats/datacats/cli/main.py", line 126, in main
    parent_exception=UndocumentedError(exc_info)))
  File "/home/ian/boxkite/git/datacats/datacats/cli/main.py", line 132, in _error_exit
    exception.pretty_print()
  File "/home/ian/boxkite/git/datacats/datacats/error.py", line 32, in pretty_print
    print self.message.format(*self.format_args)
ValueError: Single '}' encountered in format string
```

JackMc:
What was the root cause of this exception, btw?

dborzov:
uh, that is not good. looking into this

wardi:
Here's another one
```
$ datacats init
Creating environment bare/primary from existing environment directory "bare".......
Installing ckan
Installing ckan requirements
Initializing database
Starting web server at http://127.0.0.1:5565/ ...
admin user password:
confirm password:
data----------------------------------------
datacats: problem was encountered:
Traceback (most recent call last):
  File "/home/ian/boxkite/venv/bin/datacats", line 9, in <module>
    load_entry_point('datacats==1.0.0dev1', 'console_scripts', 'datacats')()
  File "/home/ian/boxkite/git/datacats/datacats/cli/main.py", line 126, in main
    parent_exception=UndocumentedError(exc_info)))
  File "/home/ian/boxkite/git/datacats/datacats/cli/main.py", line 132, in _error_exit
    exception.pretty_print()
  File "/home/ian/boxkite/git/datacats/datacats/error.py", line 32, in pretty_print
    print self.message.format(*self.format_args)
KeyError: 'self'
```

wardi:
from https://github.com/datacats/datacats/issues/193#issue-88476683

```
Initializing database
----------------------------------------
datacats: problem was encountered:
Traceback (most recent call last):
  File "/Users/dz/Envs/datacats/bin/datacats", line 9, in <module>
    load_entry_point('datacats==1.0.0dev1', 'console_scripts', 'datacats')()
[...]

PR #194, opened 2015-06-15
Hotfix: DatacatsError for undocumented exceptions failing in formatting

dborzov:
Fixes [issue#190](https://github.com/datacats/datacats/issues/190)

`.format(args)` should be called only once on `DatcatsExeptions.message`. It was not.

### DIFF
--- a/datacats/error.py
+++ b/datacats/error.py
@@ -46,10 +46,14 @@ class WebCommandError(Exception):
         self.logs = logs
 
     def __str__(self):
+        serialized_command = " ".join(self.command) \
+                if type(self.command) is list else \
+                str(self.command)
+
         return ('    Command: {0}\n'
                 '    Docker Error Log:\n'
                 '    {1}\n'
-                ).format(" ".join(self.command), self.logs)
+                ).format(serialized_command, self.logs)
 
 
 class PortAllocatedError(Exception):
--- a/datacats/error.py
+++ b/datacats/error.py
@@ -47,8 +47,8 @@ class WebCommandError(Exception):
 
     def __str__(self):
         serialized_command = " ".join(self.command) \
-                if isinstance(self.command, list) else \
-                str(self.command)
+            if isinstance(self.command, list) else \
+            str(self.command)
 
         return ('    Command: {0}\n'
                 '    Docker Error Log:\n'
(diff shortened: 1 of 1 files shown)


JackMc:
Pls fix CI

dborzov:
@JackMc right, thanks, Jack. Done

datacats/error.py
```diff
@@ -4,7 +4,10 @@
 class DatacatsError(Exception):
 
     def __init__(self, message, format_args=(), parent_exception=None):
-        self.message = message
+        try:
+            self.message = message.format(format_args)
+        except (ValueError, KeyError):
+            self.message = message
```
wardi: this is odd. Why would you be getting Value/KeyErrors it's only our code that creates these.

wardi: And shouldn't this be `message.format(*format_args)`

dborzov: You are right, it should be `*format_args`, thanks.

ValueError and KeyError are raised for this two cases correspondingly:

```python
"I like {".format(fruit=bananas) # ValueError
"I like {vegetable} ".format(fruit=bananas) # KeyError
```

datacats/error.py
```diff
@@ -4,7 +4,10 @@
 class DatacatsError(Exception):
 
     def __init__(self, message, format_args=(), parent_exception=None):
-        self.message = message
+        try:
+            self.message = message.format(*format_args)
+        except (ValueError, KeyError):
```
dborzov: You are right, it should be *format_args, thanks.

ValueError and KeyError are raised for these two cases correspondingly:

"I like {".format(fruit=bananas) # ValueError
"I like {vegetable} ".format(fruit=bananas) # KeyError

wardi: Both those cases would be our code with an error that shouldn't be silently ignored.

Better yet: let's remove format_args from the constructor and let the code that creates these instances call .format() if it wants to on the message it passes in. That way the error stays right where it was caused and unrelated code stays out of our exception classes.

dborzov: I agree. This was the original format for DatacatsError and it is really obscure and error prone. Not only parameters are passed to the exception, they are not named. So you have to count "{0}, {1}" each time and it is really obscure.

wardi:
merging, I'll make a new issue for moving string formatting out of the DatacatsError class

PR #198, opened 2015-06-15
Fix postgis

wardi:


JackMc:
Oh. Already was working on something for this! Oh well. If this works, :+1:

wardi:
@JackMc oh, you're still working? I'm fixing this by using web_command to load the scripts with one postgres container linked to the real postgres container.

wardi:
@deniszgonjanin care to kick the tires?

deniszgonjanin:
@wardi doesn't work yet with `init`. I will test with create + spatial install and let you know how that goes

deniszgonjanin:
looks like it works with `create` :+1:

wardi:
Hmm. funny that init doesn't work, it gets run from finish_init.. You could try `tweak --install-postgis` afterwards

datacats/docker.py
```diff
@@ -181,7 +182,8 @@ def web_command(command, ro=None, rw=None, links=None, stream_output=None,
         command=command,
         volumes=binds_to_volumes(binds),
         detach=False,
-        host_config=create_host_config(binds=binds))
+        host_config=create_host_config(binds=binds),
+        entrypoint=entrypoint)
```
dborzov: what is an entrypoint?

dborzov: So docker entrypoint is the script that runs when an image is used to create a container, right?

dborzov: oh, I googled it never mind: http://serverfault.com/questions/594281/how-can-i-override-cmd-when-running-a-docker-image

JackMc:
Turns out the reason that @deniszgonjanin and I are having trouble on init is that the paster command is attempting to load spatial which is then attempting to verify PostGIS's existence, which doesn't work because CKAN's db isn't initialized and therefore doesn't container PostGIS.

A workaround is to remove spatial from the plugins config, run init, add it back in, and then run install.

JackMc:
Any suggestions on how to do something like this programmatically?

wardi:
attempt at a simple fix. Now I'm not sure why create was working!

wardi:
Ok, create was working because spatial wasn't in the plugins when db init was run

deniszgonjanin:
Right, it works with create + spatial install afterwards. I will test the newest fix

deniszgonjanin:
Yay, it seems like it's working

@JackMc @wardi I haven't looked at the latest code, but if this becomes too hackish, the best solution might be to build our own postgres image and bake postgis into it from the start instead of using the official postgres image

PR #202, opened 2015-06-16
Fix docopt so that it doesn't quit silently on user error

JackMc:
The issue was that we just sys.exit when we get SystemExit (including DocoptExit). DocoptExit does magic to get the usage printed if the message is raised all the way up to __main__.


wardi:
ah, that looks better

dborzov:
:+1: wicked neat!

PR #204, opened 2015-06-16
Fix paster issue where database would change IP

JackMc:


JackMc:
@deniszgonjanin Please verify

wardi:
I can review this

JackMc:
Ok :)

wardi:
Looks good, reverts breakage introduced in 14d342c7

PR #205, opened 2015-06-16
Fix remnants of old-style DatacatsError message formatting

dborzov:
Also, sneak in the public key for peach.boxkite.ca to the list of default keys
(no risk of abuse as not authorized connections are ignored)


JackMc:
Thank you! :)

PR #206, opened 2015-06-16
wait for postgres in install_postgis script

wardi:
not pretty but it should solve the postgres db-create race


dborzov:
:+1: would you call that monkey patching? :stuck_out_tongue_closed_eyes:

wardi:
No. I might use words like "kludge" or "hack" or "here be dragons" though

JackMc:
:+1:

wardi:
ok, reproduced this bug locally. Lengthened the max wait and this change does fix the postgres race condition (if in an unpleasant way)

Issue #153, opened 2015-06-03
Make sure we don't print out random progress stuff to stdout


dborzov:
We probably should review all the basic print messages to see if they really should be there.

Running `grep -rn "print " . --include="*.py"` within `/datacats` gives the following print statements that we should give some second thought about:

```sh
./cli/create.py:57:            print "WARNING: When deploying you will need to choose a"
./cli/create.py:58:            print "target name that is at least 5 characters long"
./cli/create.py:172:            print 'At least 4 characters are required'
./cli/create.py:177:        print 'Passwords do not match'
./cli/deploy.py:46:    print "Deployed source to http://{0}.datacats.io".format(target_name)
./cli/install.py:61:        print 'ckan not found in environment directory'
./cli/pull.py:52:                print json.dumps(s)
./cli/purge.py:44:            print 'Failed to load environment.',
./cli/purge.py:45:            print 'Not deleting environment directory.'
./cli/install.py:61:        print 'ckan not found in environment directory'
./cli/less.py:27:    print 'Converting .less files to .css...'
./cli/lesscd.py:32:            print 'event! ' + str(event)  
```

wardi:
That's all under /cli/ which seems reasonable to me

JackMc:
Removed the debug line

JackMc:
@dborzov what else should be fixed to close this do you think?

dborzov:
@wardi yeah, you are right.

@JackMc which line? the "event!" one?

`./cli/install.py:61:  print 'ckan not found in environment directory'` should be reported as raised Exception message

JackMc:
@dborzov Yeah that one

PR #207, opened 2015-06-17
purge --delete-environment was leaving an venv dir

wardi:
because of an unneeded environment.fix_project_permissions() call after the datadir is already gone

### DIFF
--- a/datacats/cli/purge.py
+++ b/datacats/cli/purge.py
@@ -45,9 +45,8 @@ Default: '.'
     environment.purge_data(sites)
 
     if opts['--delete-environment']:
-        if not environment.target:
+        if environment.target:
+            rmtree(environment.target)
+        else:
             print 'Failed to load environment.',
             print 'Not deleting environment directory.'
-        else:
-            environment.fix_project_permissions()
-            rmtree(environment.target)


datacats/cli/purge.py
```diff
@@ -45,9 +45,8 @@ def purge(opts):
     environment.purge_data(sites)
 
     if opts['--delete-environment']:
-        if not environment.target:
+        if environment.target:
+            rmtree(environment.target)
+        else:
             print 'Failed to load environment.',
```
dborzov: should not this be 
```python
raise DatacatsError('Failed to load environment.')
```

wardi: quite possibly! Have we changed all the other instances of this?

dborzov: I beilieve that is the only one left. All the other cases (see [#153](https://github.com/datacats/datacats/issues/153)) has been fixed since then

dborzov: I will add it :)

wardi: My only hesitation is that this is more of a polite warning than an "error". but I guess that's ok if it's going to be printed the same way

dborzov: ok :) here it is https://github.com/datacats/datacats/pull/208

PR #213, opened 2015-06-17
[WIP] Testable tasks

wardi:
Work towards factoring out testable functions from Environment methods. Some bug fixing along the way too.

PR #214, opened 2015-06-18
Add a quiet parameter to create and install functions so that they can be used programmatically

JackMc:
As discussed in #213


wardi:
as discussed, these will probably move to Environment with a parameter for a progress update callback in the future. This is useful as-is though.

PR #222, opened 2015-06-23
Resetting the password of the admin user is now possible

JackMc:


datacats/environment.py
```diff
@@ -607,11 +608,12 @@ def create_admin_set_password(self, password):
...
-            command=['/bin/bash', '-c', '/usr/lib/ckan/bin/ckanapi '
-                     'action user_create -i -c /project/development.ini '
-                     '< /input/admin.json'],
+            command=['/scripts/update_add_admin.sh'],
             db_links=True,
-            ro={self.sitedir + '/run/admin.json': '/input/admin.json'},
+            ro={
+                UPDATE_ADD_ADMIN: '/scripts/update_add_admin.sh',
+                self.sitedir + '/run/admin.json': '/input/admin.json'
+               },
```
wardi: this can use self.user_run_script now which makes sure to run as a non-root user.

datacats/scripts/update_add_admin.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+# This expects /input/admin.json to be the generated admin.json file
+
+if /usr/lib/ckan/bin/ckanapi action user_show -c /project/development.ini id=admin; then
+    # Generate a file with info on the user and add a password w/ Python
+    /usr/lib/ckan/bin/ckanapi action user_show id=admin -c /project/development.ini > info.json
```
wardi: /tmp/info.json ? or even cooler use a pipeline :-)

JackMc: Good idea!

JackMc:
Fixed all problems pointed out, please re-CR :)

Issue #135, opened 2015-05-30
Terminology: a source directory, an environment, a site


dborzov:
datacats project has quite a few entities and naming things is hard. It is important to agree on some nomenclature as early as possible and stick to it. 

We had a discussion about this here at **DataCats.com** and here is me attempting to write up the terminology we are sticking to.

>So we have got: 
>
>- **datacats source directory** - an isolated instance of CKAN application code with custom plugins and potentially some custom CKAN tweaks made (like using a specific version of CKAN you need)
- **datacats environment** - everything that makes one working CKAN catalog website. The **environment** is nicely isolated from everything else on a docker container and is ready to be deployed. It consists of a **source directory**, an isolated python venv (a copy of python interpreter with the necessary python libraries installed), database and search engine (solr) configs. 
- **datacats site** - once you deploy and start using a **environment**, all the data you load (datasets, user profiles and so on) make up a **site**. The **environment** can run as many **sites** as needed. 
>
>It may sound a little convoluted. An analogy one can make is to compare building a data catalog website with writing a novel on your laptop.  The **site** is the novel in that scenario, the **environment** is everything stored in the memory of your laptop's hard drive (including the operating system) and **source directory** is the the machine code and your settings of your text editor.  

>The crucial thing is if there already is a **datacats environment** that fits all your needs, all you have to do is to deploy a new site off it which takes seconds with the datacats tool.  

I think we need to agree on the language for these definitions, and then update the docs and the README with them.

JackMc:
I think these are good names!

wardi:
I'd say "datacats environment" is a source directory, an installed python virtualenv and one or more datacats sites.

Each datacats site has its own database, datastore, solr index and collection of uploaded files.

PR #223, opened 2015-06-23
Terminology

dborzov:
Not the highest priority, but I still think we need to put the terminology we use right into README.md.

Related to the discussion in [#135](https://github.com/datacats/datacats/issues/135)

@JackMc , @wardi , @deniszgonjanin what do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,14 @@
 [![docs](https://readthedocs.org/projects/docs/badge/?version=latest)](http://docs.datacats.com/)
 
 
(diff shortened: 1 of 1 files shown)


wardi:
Yes, but this might be better in the proper docs than the readme

JackMc:
Yeah, we should keep the README short and sweet preferably

wardi:
closing in favour of adding this important terminology definition to the main documentation. 

I might even add a pretty diagram.

PR #225, opened 2015-06-24
Make the home directory of www-data /var/www/storage

JackMc:


datacats/environment.py
```diff
@@ -23,7 +23,8 @@
 from datacats.template import ckan_extension_template
 from datacats.scripts import (WEB, SHELL, PASTER, PASTER_CD, PURGE,
     RUN_AS_USER, INSTALL_REQS, CLEAN_VIRTUALENV, INSTALL_PACKAGE,
-    COMPILE_LESS, DATAPUSHER, INSTALL_POSTGIS, ADJUST_DEVINI)
+    COMPILE_LESS, DATAPUSHER, INSTALL_POSTGIS, ADJUST_DEVINI,
```
dborzov: *(Bikeshedding like there is no tomorrow here. Brace yourselves)*

Since we are importing so many items from this namespace and use each one only once, maybe we should start importing this explicitly with:
```
import datacats.scripts
```

In general, I am in favour of this style in most cases. It is much more explicit and improve code's readibility and comprehension.
When you encounter:
```
blah=datacats.scripts.UPDATE_ADD_ADMIN
```
it is clear right away where this thing is defined and is coming from. 

On the other hand, this:
```
blah=UPDATE_ADD_ADMIN
```
will require an extra lookup to figure out.

I think this form
```
from blah import blahblah
```
only makes sense when `blahblah` is used quite a few times (4 or more) or at least a part of something presumed really well known (like widely used parts of standard library or some key library object).

wardi: Yes, this has been bothering me for a while but not enough to fix it.

I'd merge a PR that changes these to `from datacats import scripts` and uses `scripts.FOO` instead

JackMc: I wonder why these changes are showing up in this PR... They already got merged AFAIK

wardi: the other thing I was considering was dropping the scripts module and just adding a function that returns the full path to the script so that all the repetition in scripts.py can be removed too

JackMc: +1 to this ^^ More testable

JackMc: We can even add a test to check executable permissions on scripts

datacats/scripts.py
```diff
@@ -26,3 +26,4 @@
 DATAPUSHER = SCRIPTS_DIR + '/datapusher.sh'
 INSTALL_POSTGIS = SCRIPTS_DIR + '/install_postgis.sh'
 ADJUST_DEVINI = SCRIPTS_DIR + '/adjust_devini.py'
+UPDATE_ADD_ADMIN = SCRIPTS_DIR + '/update_add_admin.sh'
```
dborzov: let's use
```
os.path.join(SCRIPTS_DIR,  'update_add_admin.sh')
```
here.

It is crazy how many stupid errors due to missing or extra "/" in the file path come from things like this.

dborzov:
:+1:  I :heart: this, Jack!

PR #226, opened 2015-06-24
New method to get scripts that doesn't involve a load of duplication

JackMc:
As discussed in comments of #225 

This is just an idea of mine, feel free to fork/CR


dborzov:
:+1:

datacats/docker.py
```diff
@@ -6,6 +6,7 @@
 
 from __future__ import absolute_import
 
+from datacats import scripts
```
wardi: I think `from datacats.scripts import get_script_path` is warranted here

datacats/docker.py
```diff
@@ -363,7 +363,9 @@ def collect_logs(name):
 
 def check_connectivity():
     c = run_container(None, 'datacats/web', '/project/check_connectivity.sh',
```
wardi: use user_run_script please (i know, not from this change...)

wardi:
otherwise looks good!

datacats/docker.py
JackMc: this is in docker.py, using user_run_script would cause a circular dep, no?

JackMc:
Closing to reopen on branch

PR #229, opened 2015-06-24
Compile bootstrap less in addition to ckan less

JackMc:


datacats/scripts/compile_less.sh
```diff
@@ -1,2 +1,9 @@
 #!/bin/bash
-lessc $1 > $2
+
+
+lessc /project/target/ckan/ckan/public/base/less/main.less \
+> /project/target/ckan/ckan/public/base/css/main.debug.css
```
wardi: should we be generating the main.css too? I think when debug is off that's the css file that's actually used

JackMc: Yes, but this is mainly a debug feature right? We should do less compiling for prod on the server side, no?

wardi: ok, so should we fix this with documentation? Tell people if they want their style changes visible on production they need to copy them to the right place? That seems like an easy way to end up with grumpy users after they run 'deploy'

JackMc: Huh, well we could just do a copy, but that would imply that the non-minified sources are good for prod :stuck_out_tongue:

PR #230, opened 2015-06-24
New method to get scripts that doesn't involve a load of duplication

JackMc:
As discussed in comments of #225


wardi:
@JackMc go ahead with the merge after conflicts are resolved

wardi:
sorry, also please `import get_script_path` to shorten "scripts.get_script_path" a bit

wardi:
@JackMc please update and merge when you have a chance

Issue #52, opened 2015-03-25
Add support for starting and linking to a redis container, for ckanext-harvest


deniszgonjanin:
It's fairly easy to do manually, but it's a nice to have for datacats to automatically run a redis container that the harvester can use for it's queue

JackMc:
Should this be another download-when-needed container?

wardi:
yes, I think so.

florianm:
+1 on redis container and any solution for `paster celeryd`!

PR #234, opened 2015-06-25
Add automatic downloading of extra containers (harvester/QA/reports support)

JackMc:
This also adds a redis container. It is not yet linked into the web container.


JackMc:
This addresses (perhaps only partly) #52

JackMc:
I would love any feedback on this

datacats/cli/manage.py
```diff
@@ -240,10 +245,11 @@ def tweak(environment, opts):
     """Commands operating on environment data
 
 Usage:
-  datacats tweak --install-postgis [ENVIRONMENT]
+  datacats tweak [--add-redis|--install-postgis] [ENVIRONMENT]
```
wardi: not quite right, tweak with no options doesn't make sense

```
Usage:
  datacats tweak --install-postgis [ENVIRONMENT]
  datacats tweak --add-redis [ENVIRONMENT]
```

datacats/environment.py
```diff
@@ -247,7 +252,7 @@ def start_supporting_containers(self):
         operate) if they aren't already running.
         """
         task.start_supporting_containers(self.sitedir, self.target,
-            self.passwords, self._get_container_name)
+            self.passwords, self.extra_containers, self._get_container_name)
```
wardi: add new parameter to the end, please

datacats/task.py
```diff
@@ -434,11 +439,16 @@ def create_source(srcdir, preload_image, datapusher=False):
             )
 
 
+# Maps container extra names to actual names
+EXTRA_IMAGE_MAPPING = {'redis': 'redis'}
+
+
 def start_supporting_containers(sitedir, srcdir, passwords,
-        get_container_name):
+        extra_containers, get_container_name):
```
wardi: er, I meant here

JackMc:
In other news, the Harvester "New" page has a problem... A required field just doesn't exist:

wardi:
is this the right version of the harvester?  If you're stuck on this, the other extensions are a bigger priority

JackMc:
I was running master - do you think that might be the problem?

JackMc:
I tried QA and it was giving me trouble only by the fact that we don't have Celery set up. It seems CKAN has some celery support built into its paster plugin, but it just seems to fail on me:

```
(ckan)shell@testenv:~$ paster --plugin=ckan celeryd
  File "/usr/lib/ckan/bin/paster", line 9, in <module>
    load_entry_point('PasteScript==1.7.5', 'console_scripts', 'paster')()
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/paste/script/command.py", line 104, in run
    invoke(command, command_name, options, args[1:])
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/paste/script/command.py", line 143, in invoke
    exit_code = runner.run(args)
  File "/usr/lib/ckan/local/lib/python2.7/site-packages/paste/script/command.py", line 238, in run
    result = self.command()
  File "/project/ckan/ckan/lib/cli.py", line 1000, in command
    self.run_()
  File "/project/ckan/ckan/lib/cli.py", line 1014, in run_
    os.environ['CKAN_CONFIG'] = os.path.abspath(self.options.config)
  File "/usr/lib/ckan/lib/python2.7/posixpath.py", line 367, in abspath
    if not isabs(path):
  File "/usr/lib/ckan/lib/python2.7/posixpath.py", line 61, in isabs
    return s.startswith('/')
AttributeError: 'NoneType' object has no attribute 'startswith'
```

JackMc:
Same issue after checking out the stable branch

deniszgonjanin:
Harvester is a generic plugin that also needs a specific harvester plugin for each 'source' to harvest from. It comes with a `ckan_harvester` plugin included. It needs to be enabled in the plugins:

`ckan.plugins = harvest ckan_harvester`

wardi:
@JackMc this is good to be reviewed?

JackMc:
Yes please

datacats/environment.py
```diff
@@ -585,6 +593,26 @@ def fully_running(self):
...
+        """
+        Add a container as a 'extra'. These are running containers which are not necessary for
+        running default CKAN but are useful for certain extensions
+        :param container: The container name to add
+        :param error_on_exists: Raise a DatacatsError if the extra container already exists.
+        """
+        if container in self.extra_containers and error_on_exists:
+            raise DatacatsError('{} is already added as an extra container.'.format(container))
+
+        self.extra_containers.append(container)
```
wardi: missing an `if container in self.extra_containers: return`

wardi:
Do harvester, QA and reports work after this change, and after the user runs the `datacats tweak --add-redis` command?

JackMc:
As long as you do the qa pull (or whatever their command is), harvester harvest and reports run command manually, yeah

wardi:
the last thing I ask for is that this tweak command has a test added to see that it does its thing properly

JackMc:
Like check that the container exists?

wardi:
Check that the command completes and that on restart the container is running, yes

deniszgonjanin:
I've been testing the redis container this morning, works great @JackMc :+1:

JackMc:
Okie, gonna add a test

JackMc:
Test added and passed!

JackMc:
:smiley:

florianm:
I just got stuck after `datacats tweak --add-redis` because I forgot I have to `datacats reload`. A hint displayed about the necessary reload when running `datacats tweak --add-redis` would be a great failsafe.

deniszgonjanin:
:+1:

Issue #232, opened 2015-06-25
add 2.4 support


wardi:
ckan 2.4 is in beta, let's start building preload-2.4 tagged images in docker and allow datacats create to make 2.4 sites.

JackMc:
I'm building a preload image on my machine now... Is there anything other than making the --ckan option work that we need to do?

JackMc:
#235 fixes this

PR #235, opened 2015-06-26
Add support for CKAN 2.4 and specifying CKAN version

JackMc:
Addresses #232


wardi:
Please add a table that indicates the branch to use for each supported ckan version, currently:

| version requested | ckan branch |
| --- | --- |
| 2.3 | release-v2.3-latest |
| 2.4 | release-v2.4.0 |

after 2.4.0 is released switch the last one to release-v2.4-latest

wardi:
this needs the lookup table I mentioned above and some tests (see that we can successfully install all supported versions)

JackMc:
Lookup tables are so haut right now

JackMc:
And I don't see how we could efficiently test the different versions?

wardi:
how about we split our tests into different scripts and use https://circleci.com/docs/parallel-manual-setup

wardi:
cool! I look forward to seeing parallel tests working. I thought we might need to add a DATACATS_DATA environment var to allow moving the default from ~/.datacats for parallel tests.

JackMc:
It seems to be currently broken and I can't really work on this particular bit atm cause CircleCI is down (outage)

JackMc:
But yes it doesn't seem too hard. I'm also working on making a docker container on which we can run the test scripts locally.

test_2.4.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+datacats create --ckan=2.4 site3
```
wardi: Do you think it's worth running any other tests across 2.3, 2.4 and master? How about the tweak commands since they "reach inside" and do things to the environment that might be broken by different ckan versions?

datacats/cli/pull.py
```diff
@@ -12,13 +12,15 @@
...
+    'datacats/ckan:2.3',
     'datacats/postgres',
     'datacats/solr'
     ]
 
-DEVEL_IMAGES = [
+EXTRA_IMAGES = [
     'datacats/lessc',
+    'datacats/ckan:latest',
+    'datacats/ckan:2.4b'
```
wardi: let's just call this '2.4' so we don't have to change it after release

JackMc: ok

JackMc:
What needs to happen with this?

Issue #237, opened 2015-06-26
verify that datacats install reloads all running sites when complete


wardi:
also check that things calling install have the same behaviour

PR #238, opened 2015-06-26
Reload all sites after install

JackMc:
Addresses #237


JackMc:
Some of that sweet sweet CR, shaken not stirred.

PR #243, opened 2015-06-29
Add a check so we don't double up the 'primary' section

### LINKED ISSUE #242: Check if port already specified in config on init

JackMc:
(or any other child section, for that matter)

Fixes #242


JackMc:
Good for CR

PR #249, opened 2015-07-06
Fix datapusher woes

JackMc:
As discussed in pull request for datapusher

### DIFF
--- a/datacats/scripts/web.sh
+++ b/datacats/scripts/web.sh
@@ -20,7 +20,7 @@ fi
 while true; do
 	# fix our development.ini
 	if [ "$2" = "True" ]; then
-		/scripts/adjust_devini.py "$(ip route get 8.8.8.8 | awk 'NR==1 {print $NF}')" "$port"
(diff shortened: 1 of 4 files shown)


datacats/scripts/web.sh
```diff
@@ -20,7 +20,7 @@ fi
 while true; do
 	# fix our development.ini
 	if [ "$2" = "True" ]; then
-		/scripts/adjust_devini.py "$(ip route get 8.8.8.8 | awk 'NR==1 {print $NF}')" "$port"
+		/scripts/adjust_devini.py "$(ip route get 8.8.8.8 | awk 'NR==1 {print $NF}')" "5000"
```
wardi: this won't work if running with apache, right?

JackMc: Durr

JackMc:
My bad, fixed in latest

wardi:
last thing we need is a regression test. Maybe just check the run/development.ini file to see if the site_url was modified properly.

JackMc:
Hmm... Thought I added that. Maybe it's on my VM

JackMc:
Test added - don't know if it passes though, as CI is down

PR #252, opened 2015-07-08
Adding proper support of log_syslog option to aggregate all the datacats container logs in host's syslog

dborzov:
## add log_syslog flag to things in datacats.task 
This adds log_syslog flag for all the container spanning functions (it was partially implemented already though, this one fixes gaps)

## turn it on for prod envs
use  log_syslog=True when Environment.always_prod=True

## assign syslog-tag
This assigns more descriptive syslog-tags when log_syslog=True.

Docker supports log-drivers: a config option for containers that enables redirecting that
container's logs somewhere else (as opposed to just dumping them in a json file
within the container's directory).

We have a flag to support syslog log-driver for new containers.
However, by default the source field of the log entry would only include container IDs
which is not very helpful when browsing the logs. Container names or something of this sort would have been much more helpful.

One can do it Docker 1.8 and higher by using the custom syslog tags feature,
where we assign the corresponding docker log option value (--log-opt). This thing uses that
and assigns container name for all cases when syslog log-driver is turned on.

### DIFF
--- a/datacats/docker.py
+++ b/datacats/docker.py
@@ -238,16 +238,20 @@ def run_container(name, image, command=None, environment=None,
     """
     Wrapper for docker create_container, start calls
 
+    :param log_syslog: bool flag to redirect container's logs to host's syslog
+
     :returns: container info dict or None if container couldn't be created
 
     Raises PortAllocatedError if container couldn't start on the
     requested port.
     """
     binds = ro_rw_to_binds(ro, rw)
-
-    host_config = create_host_config(binds=binds,
-                                     log_config=LogConfig(
-                                         type=('syslog' if log_syslog else 'json-file')))
+    log_config = LogConfig(type=LogConfig.types.JSON)
+    if log_syslog:
+        log_config = LogConfig(
+            type=LogConfig.types.SYSLOG,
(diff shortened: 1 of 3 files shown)


JackMc:
:+1: Just to verify, this works in all versions of Docker above 1.6.0?

dborzov:
Yeah, it is going to work with Docker 1.6, 1.7

JackMc:
Cool

dborzov:
@wardi I think it is ready for CR and merging :)

wardi:
This removes the --syslog option and changes a number of the Environment methods parameter lists to do the same. Instead please maintain compatibility with ckan-multisite and other code calling in to datacats by keeping that flag but just treat the always_prod case as the same as if syslog was set to True.

JackMc:
:+1:

dborzov:
@wardi fixed

JackMc:
Can you fix merge conflict?

wardi:
yeah, this is good to merge after conflicts are resolved.

dborzov:
Rebased off newer master, now waiting for tests and if all is good, merging

Issue #255, opened 2015-07-10
Docs regression


deniszgonjanin:
This command from the docs no longer appears to work: http://docs.datacats.com/faq.html#run-an-interactive-paster-shell

wardi:
Much of the FAQ is broken or obsolete. We should fix this command for sure, but I'd rather remove the FAQ and have a "recipes" or "guide" section for common things users would want to do

wardi:
fixed by #257

PR #257, opened 2015-07-11
fix shell+paster CLI parsing and add unit tests

wardi:
:grin:


wardi:
fixes #255

JackMc:
:fearful: is the world ending

wardi:
@JackMc no, it just seems that way because datacats has the start of some useful unit tests now

JackMc:
:wink: I like the sass

deniszgonjanin:
@JackMc Yeezy not rocking your world?

Issue #253, opened 2015-07-09
How to set ckan.site_url


florianm:
I'm trying to enable CKAN extensions (detailed install steps see my [write-up](http://govhack2015.readthedocs.org/en/latest/)) and have set the ckan.site_url, which for some reason does seem to be ignored.

FWIW, I've installed datacats into a venv (to avoid pip/requests conflicts), and hooked up my AWS VM through Route53 / ElasticIP to my domain ([write-up](http://govhack2015.readthedocs.org/en/latest/1_DevOps/)).
My setup uses nginx to reverse proxy local ports to subdomains on port 80, so my `ckan.site_url` is e.g. http://datacats.yes-we-ckan.org. At a stretch I could use the site's IP (which is the AWS VM's private address - I'd prefer to use the static IP set by ElasticIP), but the port (80) is definitely not the local datacats environment's port (in my example 5687).

E.g. a the spatial preview of an example [dataset](http://datacats.yes-we-ckan.org/dataset/example-dataset) is blank with a `data-module-site_url=""http://127.0.0.1:5687/""`

Also, the resource url is 127.0.0.1/... rather than my site_url. This of course makes the resource proxy [unhappy](http://datacats.yes-we-ckan.org/dataset/example-dataset/resource/3aedf8cd-1a23-4179-8b26-68fe137548f5).

I understood that [adjust_devini.py](https://github.com/datacats/datacats/blob/master/datacats/scripts/adjust_devini.py) overrides the site_url with the IP consumed by [web.sh](https://github.com/datacats/datacats/blob/76b13a2ca9ad886114df081161b82039daacb738/datacats/scripts/web.sh#L23) (that line works fine on my end and returns my public IP), but even that IP (my public IP) doesn't seem to make it.

Could the site_url be set in order of priority:

* from ckan config's `ckan.site_url` (not working on my end),
* or inferred from current IP lookup as per web.sh (not working),
* or defaulting to 127.0.0.1 (works).

It would be awesome to have a short section in the docs demonstrating how to set the site_url if it has to be set other than through the development.ini setting, or if the setting could be used over any defaults.

I've tried the `--address=IP` option to no avail.

JackMc:
If you wish to explicitly set the site_url while we're working on a command line option for it, you can edit the ``.datacats-environment`` file. It's a simple INI

florianm:
Yes, I need to set the site_url, otherwise my resource_proxy, all previews and the datapusher won't work.
[...]

PR #258, opened 2015-07-13
Add an option to specify the site_url

JackMc:
Since the port is (typically) determined by Datacats at runtime, along with sometimes the address, we provide templating on the site_url through the Python format() syntax.

Addresses #253. (Hopefully)

### DIFF
--- a/datacats/cli/manage.py
+++ b/datacats/cli/manage.py
@@ -54,7 +54,7 @@ Options:
   -s --site=NAME        Specify a site to start [default: primary]
   --syslog              Log to the syslog
   --site-url SITE_URL   The site_url to use in API responses. Defaults to old setting or
-                        will attempt to determine it. (e.g. http://myrandomhostname.org:{port}/)
+                        will attempt to determine it. (e.g. http://example.org:{port}/)
 
 ENVIRONMENT may be an environment name or a path to an environment directory.
 Default: '.'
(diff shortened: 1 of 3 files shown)


datacats/cli/manage.py
```diff
@@ -70,28 +72,41 @@ def reload_(environment, opts):
...
+    if opts['PORT'] or opts['--address'] != '127.0.0.1' or opts['--site-url']:
         if opts['PORT']:
             environment.port = int(opts['PORT'])
         if opts['--address'] != '127.0.0.1':
             environment.address = opts['--address']
+        if opts['--site-url']:
+            site_url = opts['--site-url']
+            # TODO: Check it against a regex or use urlparse
+            try:
+                site_url = site_url.format(address=environment.address, port=environment.port)
```
florianm: If `opts["--site-url"]` is specified, doesn't this bit overwrite it with 127.0.0.1 and the given port?

JackMc: Nopers, this is just so that users who need to know the port can use it in specifying their site url like this:

datacats create mysite --site-url "http://my-random-site.com:{port}/"

(Datacats will substitute the correct port in for "port" here

JackMc: The problem is most likely that datacats is detecting that your "address" variable is still set to 127.0.0.1, and trying to fix that by changing around the site_url.

JackMc:
Ok, @florianm can you try this new change? We are now a little bit more restrictive about generating the site_url from the docker IP (we only do it if there's no specified site_url)

wardi:
@JackMc please merge after resolving conflict

Issue #248, opened 2015-07-02
development.ini owned by root after datacats create


wardi:
we should be running the ini-creation as the shell user

wardi:
this is happening for 1.0 too :-/

JackMc:
Backport to 1.0?

wardi:
already done. I'm looking for other small fixes that might go in to the point release

PR #260, opened 2015-07-14
Run make-config as user so we don't have development.ini owned by root

JackMc:
Addresses #248


wardi:
@JackMc manual rebuild? I can't seem to trigger it because it's on your fork

JackMc:
Working now

Issue #261, opened 2015-07-14
retry pull a few times


wardi:
docker hub seems gassy at times, give it a chance to finish by retrying failed pulls before giving up

JackMc:
Okie dokie

PR #263, opened 2015-07-14
Retry pull when it fails

JackMc:
Addresses #261


JackMc:
Reopening on a better branch

PR #264, opened 2015-07-14
Retry pull when it fails

JackMc:
Addresses #261

### DIFF
--- a/datacats/cli/pull.py
+++ b/datacats/cli/pull.py
@@ -38,7 +38,22 @@ Options:
                      you are going offline.
 """
     for i in IMAGES + (EXTRA_IMAGES if opts['--all'] else []):
-        pull_image(i)
+        retrying_pull_image(i)
+
+
+def retrying_pull_image(image_name):
+    while True:
+        retries = 0
+        try:
+            retries += 1
+            pull_image(image_name)
+            break
+        except DatacatsError:
+            if retries <= 5:
+                print
+                print 'Error while pulling image {}. Retrying.'
+            else:
+                raise
 
 
 def pull_image(image_name):
--- a/datacats/cli/pull.py
+++ b/datacats/cli/pull.py
@@ -65,8 +65,8 @@ def _retry_func(func, param, num, retry_notif, error_msg):
 
     Throws DatacatsError if we run out of retries. Returns otherwise.
     """
-    for retry_num in range(1, num + 1):
-        if retry_num - 1:
+    for retry_num in range(num):
+        if retry_num:
             retry_notif(param, retry_num)
         try:
             func(param)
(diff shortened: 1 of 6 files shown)


datacats/cli/pull.py
```diff
@@ -38,7 +38,22 @@ def pull(opts):
...
+    while True:
+        retries = 0
+        try:
+            retries += 1
+            pull_image(image_name)
+            break
+        except DatacatsError:
+            if retries <= 5:
+                print
+                print 'Error while pulling image {}. Retrying.'
```
wardi: did you forget a .format() here?

datacats/cli/pull.py
```diff
@@ -38,7 +38,22 @@ def pull(opts):
                      you are going offline.
 """
     for i in IMAGES + (EXTRA_IMAGES if opts['--all'] else []):
-        pull_image(i)
+        retrying_pull_image(i)
+
+
+def retrying_pull_image(image_name):
+    while True:
```
wardi:
```python
for retries in range(1, 6):
```
would let you remove the `retries = 0` and `retries += 1` and it also has the benefit of actually stopping after 5 retries ;-)

JackMc:
What bugs? :)

JackMc:
There's no bugs! Just features.

datacats/cli/pull.py
```diff
@@ -38,7 +38,20 @@ def pull(opts):
...
+    for retries in range(1, 6):
+        try:
+            pull_image(image_name)
+            break
+        except DatacatsError:
+            if retries <= 5:
+                print
+                print 'Error while pulling image {}. Retrying.'.format(image_name)
+            else:
+                raise
```
wardi: still wrong, I think. This will fail silently because retries never reaches the value 6. Maybe try writing this as a function that you can add a test for to see if it actually works

JackMc: Will do

JackMc:
Test added, passes.

datacats/tests/test_cli_pull.py
```diff
@@ -0,0 +1,23 @@
...
+        count.counter = 0
+
+        try:
+            _retry_func(raise_an_error, None, 5, count,
+                        'Error! We wanted this to happen')
+            self.fail('Exception was not raised.')
+        except DatacatsError:
+            pass
+        finally:
+            self.assertEqual(count.counter, 5)
```
wardi: Hmm. does fail raise an exception? Might be better to move this assertEqual into the except statement. Also, let's check that the error that was raised has 'We wanted this to happen' in it, so our pass criteria is a little tighter.

datacats/cli/pull.py
```diff
@@ -38,7 +43,36 @@ def pull(opts):
...
+    :param: error_msg: The message
+
+    Throws DatacatsError if we run out of retries. Returns otherwise.
+    """
+    for retry_num in range(1, num + 1):
+        try:
+            func(param)
+            return
+        except DatacatsError:
+            retry_notif(param, retry_num)
```
wardi: as written this will write "retrying" right before failing.  How about this at the top of the loop instead
```python
for retry_num in range(num):
    if retry_num: retry_notif(param, retry_num)
    ...
```

Issue #241, opened 2015-06-29
document clear manual steps for cleaning a directory


deniszgonjanin:
Related to #240 

If `datacats migrate` fails, no datacats commands can be used on the environment. This includes `purge`. To the user, it is not clear how to recover from `migrate` failing.

Ideally `purge` should always work regardless of the format version. If that is too difficult, we should print out a message with some instructions for cleaning the environment manually when `datacats purge` fails.

JackMc:
:+1:

JackMc:
I think it's a better idea for us to just fallback to removing containers for the old format. More user-friendly.

JackMc:
#268 takes a stab at fixing things

PR #268, opened 2015-07-17
Add support for format version detection to purge

JackMc:
Addresses #241


wardi:
Please add a test and merge from master to fix circleci

wardi:
@JackMc this is the last thing I'd like for 1.1.0, would you add the test when you have a minute?

JackMc:
Huh, weird. Thought I'd added a test for this.

wardi:
getting closer!

PR #272, opened 2015-07-21
Cli progress bar and generilized progress tracking for datacats commands

dborzov:
Potentially it is possible to call the datacats commands in environments beyond the interactive cli. For example, commands can be put in a Celery queue .

For these reasons, it would help to generalize the progress tracking to support other running contexts. At the moment the commands report their progress just by printing dots and progress messages to `stderr`. A transition is needed to a more generic approach.

This PR suggest one. 

We add `progress_tracker` key argument to our datacats command implementations. If none passed, by default, we use the CliProgressTracker that generates a nice-looking tty progress bar for the user. 

However, it is also easy to make and use custom progress tracker as well. The commands only call
the `update_state` method to communicate their stage. It goes like this:
```python
progress_tracker.update_state(state='PROGRESS',
                    meta={
                        'current': 20,
                        'total': 100,
                        'status': 'Making sandwiches'}
     )
                    
```
note  that this interface is identical to the one Celery uses.

### DIFF
--- a/datacats/cli/create.py
+++ b/datacats/cli/create.py
@@ -12,7 +12,7 @@ from datacats.environment import Environment
 from datacats.cli.install import install_all
 from datacats.error import DatacatsError
 
-from datacats.cli.util import CliProgressTracker, y_or_n_prompt, confirm_password, function_as_step
+from datacats.cli.util import CLIProgressTracker, y_or_n_prompt, confirm_password, function_as_step
 
 
 def write(s):
@@ -93,7 +93,7 @@ def create_environment(environment_dir, port, ckan_version, create_skin,
 
         if not progress_tracker:
             # by default we use a cli progress tracker
-            progress_tracker = CliProgressTracker(
+            progress_tracker = CLIProgressTracker(
                 task_title='Creating datacats site environment',
                 total=len(steps),
                 quiet=quiet)
--- a/datacats/cli/create.py
+++ b/datacats/cli/create.py
@@ -5,13 +5,14 @@
 # See LICENSE.txt or http://www.fsf.org/licensing/licenses/agpl-3.0.html
 
 import sys
+import types
 from os.path import abspath
 
 from datacats.environment import Environment
 from datacats.cli.install import install_all
 from datacats.error import DatacatsError
 
-from datacats.cli.util import y_or_n_prompt, confirm_password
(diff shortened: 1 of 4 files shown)


dborzov:
@wardi , @JackMc do you guys think that is a good approach to progress tracking?

datacats/cli/create.py
```diff
@@ -5,13 +5,14 @@
...
 import sys
+import types
 from os.path import abspath
 
 from datacats.environment import Environment
 from datacats.cli.install import install_all
 from datacats.error import DatacatsError
 
-from datacats.cli.util import y_or_n_prompt, confirm_password
+from datacats.cli.util import CliProgressTracker, y_or_n_prompt, confirm_password, function_as_step
```
wardi: CLIProgressTracker is a better name, according to pep8 :stuck_out_tongue:

datacats/cli/create.py
```diff
@@ -65,43 +66,58 @@ def create_environment(environment_dir, port, ckan_version, create_skin,
...
+            steps.append(function_as_step(environment.create_install_template_skin))
+
+        if not progress_tracker:
+            # by default we use a cli progress tracker
+            progress_tracker = CliProgressTracker(
+                task_title='Creating datacats site environment',
+                total=len(steps),
+                quiet=quiet)
+        for step_num, step in enumerate(steps):
+            if not isinstance(step, types.FunctionType):
```
wardi: with your function_as_step this isinstance shouldn't be required, right?

wardi:
This is in the right direction, but if we're fixing the interface I would prefer to move this whole block into the Environment class' new method at the same time. It's silly for the interface we're building for external users to be in a module under datacats.cli.

wardi:
@dborzov just repeating what we talked about
* create a new datacats/commands.py as a module to support 
* move the cli/create.py changes to datacats/commands.py
* make the list of steps a simple list of functions to call or (function to call, dict of parameters) tuples and always take the step description from the function docstring

JackMc:
Maybe do a ``from datacats.commands import *`` in the __init__.py to make this clear that this is the supported way to interact with the library?

dborzov:
@JackMc , yeah, that was Ian's thinking too

wardi:
closing in favour of the newer work on #273

PR #273, opened 2015-07-22
Refactor datacats so that it can be used as a package

dborzov:
datacats has been made as a cli-tool. We now see that it can be extensively used as python package and the interface should be the sam as the datacats-cli. See discussion here for context: https://github.com/datacats/datacats/pull/272

This PR scaffolds this Le Refactoree Grandiouse:

* the create command has been properly moved to commands.py
* install_all and finish_init were moved to task.py as they don't belong to cli package

I am making a PR now so that you guys @wardi @JackMc could see it and provide the feedback on this approach. Once we settle on this, refactoring all the other commands should be straightforward enough.

### DIFF
--- a/datacats/cli/create.py
+++ b/datacats/cli/create.py
@@ -8,12 +8,10 @@ import sys
 import types
 from os.path import abspath
 
-from datacats.environment import Environment
-from datacats.cli.install import install_all
+import datacats
 from datacats.error import DatacatsError
 
-from datacats.cli.util import CLIProgressTracker, y_or_n_prompt, confirm_password, function_as_step
-
+from datacats.cli.util import CLIProgressTracker, y_or_n_prompt, confirm_password
 
 def write(s):
     sys.stdout.write(s)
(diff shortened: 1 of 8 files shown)


JackMc:
Does this work if progress_tracker == None?

dborzov:
@JackMc yes it does. `progress_tracker` being None is the default and it means it is silent by default as packages should be. CLI wrappers from the cli commands would initiate their own instance of `CLIProgressTracker` and pass it on when calling the very same commands from the same interface.

datacats/cli/create.py
```diff
@@ -223,43 +163,3 @@ def init(opts, no_install=False, quiet=False):
     return finish_init(environment, start_web, create_sysadmin, address,
```
JackMc: Isn't this broken now this function isn't in this module?

JackMc:
Init needs one of those command thingies

datacats/cli/create.py
dborzov: yes it is broken now. this is for reset command which I have not touched yet.

dborzov:
@JackMc what thingies

JackMc:
Nvm.

Responding to your question in OP, the only problem I can see here is that the task module is gonna become bigger and bigger and kinda defeat the purpose of having a module for taking stuff out of environment

dborzov:
@JackMc that is not the purpose of this refactoring. The purpose is to separate out and define the public/interface functions once and for all so that we don't have to change all the calls every time there is another refactor for the projects that use datacats as a package. Like ckan-multisite.

wardi:
finish_init and install_all should be in commands, not task. They are things that encompas many steps, need a progress bar and an environment object.

task.py is set up to be small easily testable functions that don't take the environment object (just simple parameters)

wardi:
+ their steps should be included in the things that call them (init and create)

wardi:
closing in favour of #316

PR #275, opened 2015-07-22
Add support for specifying you want SSL

JackMc:
As discussed


wardi:
The two env vars that are used by routes to create the fully qualified urls are HTTP_X_FORWARDED_PROTO and HTTP_X_FORWARDED_HOST. It makes more sense to set both of them based on any overridden site_url setting.

JackMc:
No need for this anymore

PR #280, opened 2015-07-27
Make paster command work with zero arguments (print out help)

JackMc:
It would cause an IndexError otherwise.


wardi:
doesn't paster's help just come up if you run it without a command?

JackMc:
You get an assertionerror

JackMc:
But yeah that might be better UX

wardi:
paster's help screen would be more helpful than ours :-)

JackMc:
All sorted.

Issue #281, opened 2015-07-28
datacats reset & datacats install should strip .pyc files from environment


deniszgonjanin:
For myself, a common use case for `datacats reset` is when I am switching between CKAN versions and I need a clean database.

If there are old .pyc files around, the database [migrations fail](http://docs.ckan.org/en/ckan-1.6/common-error-messages.html#assertionerror-there-is-no-script-for-46-version) when downgrading from a newer CKAN version to the old one.

I don't think there is any downside to cleaning out these files. Can they be removed automatically?

JackMc:
Hmm, they probably could. Would it be useful if we did it for all installed extensions also?

deniszgonjanin:
The extensions haven't caused me trouble, but if we're doing ckan/ let's do everything else as well

PR #284, opened 2015-07-31
Clean pyc files on clean install, and on reset.

JackMc:
Addresses #281

### DIFF
--- a/datacats/cli/create.py
+++ b/datacats/cli/create.py
@@ -8,7 +8,7 @@ import sys
 from os.path import abspath
 
 from datacats.environment import Environment
-from datacats.cli.install import install_all
+from datacats.cli.install import install_all, clean_pyc
 from datacats.error import DatacatsError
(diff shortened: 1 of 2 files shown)


datacats/cli/create.py
```diff
@@ -124,6 +124,7 @@ def reset(environment, opts):
     print 'Resetting...'
     environment.stop_supporting_containers()
     environment.stop_ckan()
+    clean_pyc(environment)
```
wardi: reset affects the database, I don't see how cleaning the pyc applies

JackMc: Denis wanted it because of CKAN upgrades.

wardi: @deniszgonjanin don't you have to `install` when you upgrade ckan?

JackMc:
Waiting until release-time for this to be merged.

PR #285, opened 2015-07-31
Fix video url in docs

JackMc:


docs/index.rst
```diff
@@ -15,7 +15,7 @@ CKAN development & deployment with Docker
     <div class="row" style="margin-bottom:20px;">
         <div class="columns small-12 medium-8 medium-offset-2" style="text-align: center;">
             <div class="flowplayer">
-                <video src="http://www.datacats.com/media/datacats_intro.mp4" width=665 height=416></video>
+                <video src="https://www.datacats.com/media/datacats_intro.mp4" width=665 height=416></video>
```
wardi: are those flowplayer links above correct?

JackMc: They seem to work, so I can only guess they are.

wardi: they're ignored if you're viewing the html5 video directly, I'll try to test on an old browser at some point

Issue #277, opened 2015-07-24
reload --interactive


wardi:
For great debugging justice!

PR #286, opened 2015-08-04
Add an 'interactive' option to the commands which start a web container.

JackMc:
Addresses #277

### DIFF
--- a/datacats/cli/create.py
+++ b/datacats/cli/create.py
@@ -146,7 +146,7 @@ def init(opts, no_install=False, quiet=False):
     """Initialize a purged environment or copied environment directory
 
 Usage:
-  datacats init [-in] [--syslog] [-s NAME] [--address=IP]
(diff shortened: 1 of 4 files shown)


deniszgonjanin:
great! I will give this a shot

JackMc:
There's a couple issues with this that are merge-blocking:
- [ ] Sometimes it will still check for connectivity after the CKAN process is ended (always failing)
- [ ] You cannot simply control-C out of the process, you have to close the window or killall it

deniszgonjanin:
killall is how I end all my processes

JackMc:
I'm gonna imagine that was written in the style of the Dos Equis guy.

Like this

deniszgonjanin:
The CKAN site itself also is unreachable after a `datacats reload -i`, though the process seems to have started fine and the container ports are open.

wardi:
closing in favour of #316

Issue #295, opened 2015-08-11
`datacats reset` reassigns port in .datacats-environment


deniszgonjanin:
I don't know if this is desired behavior. The particular environment was working fine with the port set manually. But on running `datacats reset`, the port is reassigned to what datacats feels it should be based on the env name.

This seems like a side-effect that could throw a user for a real spin trying to figure out why the site just changed ports

JackMc:
Whoops, will fix this now!

PR #296, opened 2015-08-12
Save port before purging on reset

JackMc:
Addresses #295 

@deniszgonjanin Can you take a look at this and confirm if it fixes the problem?


JackMc:
(Works on my machine ™)

deniszgonjanin:
will do

Issue #299, opened 2015-08-14
Webserver started at different ip address after reload


makkus:
So, when I create a datacats instance using the --address flag, the webserver seems to be listening on this interface correctly:

    $ datacats create --address=172.17.42.1 uoa   
    Creating environment "uoa/primary"...........
    Installing ckan
    Installing ckanext-uoatheme
    Installing datapusher
    Installing ckan requirements
    Installing datapusher requirements
    Initializing database
    Starting web server at http://172.17.42.1:5859/ ...
    admin user password:
    confirm password:

But when I do a 'datacats reload', it falls back to listening on 127.0.0.1:

    $ cd uoa                                   
    $ datacats reload
    Starting web server at http://127.0.0.1:5859/ ..

This breaks the datapusher, and also access to the resource. I'm using a linux host and run datacats directly, if that makes any difference.

JackMc:
Hi there,

That does seem to make a difference, I saw this
this afternoon. What version of datacats are you
running?

makkus:
$ datacats --version
1.0.3

JackMc:
Okie dokie, working on this now :)

PR #302, opened 2015-08-18
Do not provide 'address' parameter to start_ckan

JackMc:
remove this parameter as it's confusing.

Addresses #299

### DIFF
--- a/datacats/cli/create.py
+++ b/datacats/cli/create.py
@@ -168,7 +168,7 @@ ENVIRONMENT_DIR is an existing datacats environment directory. Defaults to '.'
     log_syslog = opts['--syslog']
 
     environment = Environment.load(environment_dir, site_name)
-    if opts['--address']:
+    if address:
         environment.address = address
     if port:
         environment.port = int(port)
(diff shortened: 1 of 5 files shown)


datacats/cli/create.py
```diff
@@ -168,7 +168,8 @@ def init(opts, no_install=False, quiet=False):
     log_syslog = opts['--syslog']
 
     environment = Environment.load(environment_dir, site_name)
-    environment.address = address
+    if opts['--address'] != '127.0.0.1':
```
wardi: `if address != '127.0.0.1':` helps me see what this code is doing better

datacats/cli/create.py
```diff
@@ -206,12 +207,12 @@ def init(opts, no_install=False, quiet=False):
...
         raise
 
-    return finish_init(environment, start_web, create_sysadmin, address,
+    return finish_init(environment, start_web, create_sysadmin,
                        log_syslog=log_syslog, do_install=not no_install,
                        quiet=quiet, site_url=site_url)
 
 
-def finish_init(environment, start_web, create_sysadmin, address, log_syslog=False,
+def finish_init(environment, start_web, create_sysadmin, log_syslog=False,
```
wardi: Is this function called from places outside of datacats?

JackMc: As far as I know, no. Don't think it is in multisite. I will check

Issue #301, opened 2015-08-18
pycsw example


florianm:
Is there a working example for these two use cases:

* expose datacats CKAN as CSW using pycsw
* expose a remote WMS as CSW using pycsw

Could a pycsw server be run as a container like celery? 

Assuming I've installed pycsw like any other extension by cloning it into the datacats environment, and have installed the requirements:
```
sudo apt-get install -y python-dev libxml2-dev libxslt-dev libgeos-dev
pip install sqlalchemy
```

The vanilla way `python pycsw/bin/pycsw-admin.py -c post_xml -u http://localhost:PORT/pycsw/csw.py -x pycsw/post.xml` indicates that binary `pycsw/bin/pycsw-admin.py` and config `pycsw/post.xml` live inside pycsw in my datacats environment. What would be recommended settings for:

* database - sqlite or postgres (do I need to use the credentials of the postgres container or spin up another db container)
* log file location
* pycsw server home

Is there a "datacats" way or should I create supervisor jobs running my pycsw servers?

JackMc:
You can submit a pull request adding ``libgeos-dev`` to the Dockerfile's list of installs, and then running it with ``datacats shell``.

deniszgonjanin:
This example is interesting because pycsw is a standalone piece of software, with it's own processes, database connections, etc... The approach above might work fine, but now we're trying to shoehorn another application into the datacats way of doing things, which is sometimes very CKAN-centric.

This is something we have talked about, but we haven't reached any agreement on what the best strategy is regarding supporting this. Anything except supporting CKAN only leads us down a path of trying to build a general-purpose web application devops tool. Which would be really cool, but sounds like a lot of work.

A PR for `libgeos-dev` to the Dockerfile is probably a good idea anyway.

florianm:
Also, I found a pycsw docker image which runs nicely. Once I've sorted out
the proxy/firewall issues I think I can provide a working example.

florianm:
nearly working... 
after I ran `datacats tweak --add-pycsw` pycsw got registered the datacats/pycsw container in envmane/.datacats-environment. So far so good!

```
datacats reload
...
APIError: 500 Server Error: Internal Server Error ("Cannot start container a6d77ebc871dfcbf836aa97e632c196290f90d7ea6b98bc17ef4ba9203282b14: Cannot link to a non running container: /datacats_pycsw_landgate_primary AS /datacats_web_landgate_primary/pycsw")
```
[...]

PR #303, opened 2015-08-19
#301 Pycsw docker image

florianm:
Here's an initial docker image for pycsw.

## Features

* base image from datacats/web
* additional steps based on https://github.com/GeographicaGS/Docker-PyCSW
* blatantly hooked into ./build_images.sh

## Next steps

* write log file to local dir, or provide a log command `pycsw log` working like `datacats log`
* copy local config to `projects/config` instead of baking it into the image so users can change the pycsw catalogue metadata settings like their contact details
* `datacats tweak --add-pycsw` should create the local folders, spin up the pycsw container (localhost:9000 maybe?), spit out a short usage description
* `datacats start` and `datacats reload` should start/restart the container


JackMc:
Wow! This is really cool. I can't wait to try it out.

JackMc:
So for this to be considered part of DataCats, I'd say, we should probably have it such that you can add arbitrary containers as extra containers, or as you say, have a ``tweak --add-pycsw`` option.

deniszgonjanin:
the `tweak -add` thing could be a good idea, esp. since it's being used for redis. We may in the future want it for other things too, so we could think about defining a more formal interface for creating these endpoints.

JackMc:
I already have code somewhere in a discarded branch that does that - let me see if I can find it

JackMc:
https://github.com/JackMc/datacats/commit/ef4e3e2b344b796ff30f619a4492e77b4038f089 This is the commit where I did that.

JackMc:
(Note that there'd probably have to be some culling of the diff there because it has some geopusher stuff in there (which instead of including we rewrote the extension))

florianm:
@JackMc I tried to add the management hooks but not sure whether I got it all right. Thanks for your diff, it made me understand the container name lookup.

My main issue atm is that this will spin up one pycsw container. What if I have, say, five WMS endpoints to harvest - wouldn't I need five pycsw containers to harvest them in? Or could I harvest five WMS endpoints into one pycsw container, and harvest that one into CKAN? I'm still stuck with validation errors due to a mismatch of WMS / pycsw / CKAN data schemas (might need to customise the CSW harvester), so I can't yet tell whether one pycsw server will be enough per ckan instance.

Issue #309, opened 2015-08-26
boot2docker address setting needs better default


deniszgonjanin:
```
(datacats)dzs-MacBook-Pro:dcats-envs dz$ datacats create testaddress
Creating environment "testaddress/primary"...........
Installing ckan
Installing ckanext-testaddresstheme
Installing datapusher
Installing ckan requirements
Installing datapusher requirements
Initializing database

----------------------------------------
datacats: problem was encountered:
Cannot specify address on boot2docker.
----------------------------------------
```

`.datacats-environment` doesn't set an address by default anymore. `datacats` itself defaults the address to `127.0.0.1`, but then raises an error for said address.

Setting `address` to the boot2docker ip works well. 

A possibly related problem is that I think previous versions of datacats did write `address = 127.0.0.1` into the `.datacats-environment` file, so those may be broken on OSX when users upgrade.

wardi:
Shouldn't address be ignored always when using boot2docker?

JackMc:
Yeah, but it wasn't being due to the == instead of !=

wardi:
No, I mean if an address is specified in the config it's not the same as a user passing it on the command line. We should ignore any address value set in the .datacats-environment file with boot2docker

JackMc:
We basically do that already, except we error if they specify anything other than 127.0.0.1 in there

JackMc:
The reason for the ignoring 127.0.0.1 is backwards compat

JackMc:
I think this is fixed?

PR #310, opened 2015-08-26
Fix boot2docker support through introduction of inverted equality symbol

JackMc:
Hot fix to address #309


deniszgonjanin:
genius!

wardi:
@JackMc please ping me when this is ignoring all address settings form the config on boot2docker

JackMc:
Tested & ready for review.

PR #315, opened 2015-09-03
Better Error messages for when Docker connection fails

JackMc:


wardi:
@JackMc is this out of date now, with the references to boot2docker?

JackMc:
Definitely is! Will work on it

JackMc:
'Tis! I'll look into fixing it :-)

JackMc:
Waiting to see if CI will pass. If it does, then this will make errors like #355 much easier for users to solve

PR #317, opened 2015-09-04
Fix checking to support both docker-machine and boot2docker.

JackMc:


datacats/docker.py
```diff
@@ -24,7 +23,7 @@
 from docker import Client
 from docker.constants import DEFAULT_DOCKER_API_VERSION
 from docker.utils import kwargs_from_env, compare_version, create_host_config, LogConfig
-from docker.errors import APIError
+from docker.errors import APIError, TLSParameterError
```
wardi: new imports from docker mean we might need to update the required version of docker-py, please check that

datacats/docker.py
```diff
@@ -42,44 +41,59 @@ def rev_cmp(a, b):
...
+try:
+    _docker_kwargs = kwargs_from_env()
+except TLSParameterError:
+    print ('Please create your docker-machine VM with the command'
+           ' "docker-machine create --driver=virtualbox dev"')
+    exit(1)
 
 
 def _get_docker():
     global _docker
```
wardi: can I get a sys.platform test that skips all this on linux?

datacats/docker.py
JackMc: Still exists in our latest supported version. All good!

Issue #323, opened 2015-09-08
`--no-initdb` option on `datacats reset`?


deniszgonjanin:
When loading an existing database into a new datacats instance, it could be useful to skip the db init step, so that a postgres dump can be loaded without having to go through the extra step of doing `paster db clean`, which itself needs a [workaround](https://github.com/ckan/ckan/issues/2306) to work with datacats.

JackMc:
Sounds like a good idea!

JackMc:
Are you working on this or can I?

deniszgonjanin:
it's all yours if you want it

JackMc:
Sounds good :3

deniszgonjanin:
Let's add the same to `datacats init` :smile:

JackMc:
Whoops, I did some code for this and forgot to PR it

JackMc:
Let me grab it

JackMc:
(See #324)

PR #324, opened 2015-09-10
Add option --no-initdb to create and init for easier importing of other CKANs

JackMc:
Addresses #323

### DIFF
--- a/datacats/cli/create.py
+++ b/datacats/cli/create.py
@@ -25,16 +25,18 @@ def create(opts):
 
 Usage:
   datacats create [-bin] [-s NAME] [--address=IP] [--syslog] [--ckan=CKAN_VERSION]
-                  [--no-datapusher] [--site-url SITE_URL] ENVIRONMENT_DIR [PORT]
+                  [--no-datapusher] [--site-url SITE_URL] [--no-init-db] ENVIRONMENT_DIR
+                  [PORT]
 
 Options:
   --address=IP            Address to listen on (Linux-only)
   --ckan=CKAN_VERSION     Use CKAN version CKAN_VERSION [default: 2.3]
   -b --bare               Bare CKAN site with no example extension
   -i --image-only         Create the environment but don't start containers
+  --no-datapusher         Don't install/enable ckanext-datapusher
+  --no-init-db            Don't initialize the database. Useful for importing CKANs.
   -n --no-sysadmin        Don't prompt for an initial sysadmin user account
   -s --site=NAME          Pick a site to create [default: primary]
-  --no-datapusher         Don't install/enable ckanext-datapusher
   --site-url SITE_URL     The site_url to use in API responses (e.g. http://example.org:{port}/)
   --syslog                Log to the syslog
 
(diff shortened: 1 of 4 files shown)


wardi:
This new switch should force --image-only

wardi:
or maybe require that the user also specifies --image-only? That might make it clearer

```
datacats create [-bn] [-s NAME] [--address=IP] [--syslog] [--ckan=CKAN_VERSION]
                  [--no-datapusher] [--site-url SITE_URL] [-i [--no-init-db]]
                  ENVIRONMENT_DIR [PORT]
```

deniszgonjanin:
The command works beautifully, but running paster commands afterwards gives me:
```
FATA[0000] Error response from daemon: Could not get container for datacats_datapusher_bcgov_primary
```

Is there something in datacats that relies on datapusher running when running a datacats shell/paster command is ran?

JackMc:
I'm guessing what's going on is that data pusher is failing cause the database is broken. Try with --no-datapusher.

deniszgonjanin:
yes, it works well when env is created without datapusher. Unfortunately we can't have a --no-datapusher command on any command except create, because datapusher is automatically on when the directory is there. So `datacats init --no-init-db` will break for most datacats instances.

Any ideas on how we can easily have `datacats init --no-init-db --no-datapusher` work? I think the --no-init-db command is really more useful on existing or pre-packages datacats environments

JackMc:
We can make the detection logic more complicated for data pusher (i.e. have a flag in the Env class)

JackMc:
@wardi It already does force --image-only and --no-sysadmin

datacats/cli/create.py
```diff
@@ -25,16 +25,18 @@ def create(opts):
...
+                  [--no-datapusher] [--site-url SITE_URL] [--no-init-db] ENVIRONMENT_DIR
+                  [PORT]
 
 Options:
   --address=IP            Address to listen on (Linux-only)
   --ckan=CKAN_VERSION     Use CKAN version CKAN_VERSION [default: 2.3]
   -b --bare               Bare CKAN site with no example extension
   -i --image-only         Create the environment but don't start containers
+  --no-datapusher         Don't install/enable ckanext-datapusher
+  --no-init-db            Don't initialize the database. Useful for importing CKANs.
```
wardi:
```
--no-init-db            Advanced use: database must be manually
                        initialized before site will start. Implies
                        --no-datapusher and --image-only options.
```

datacats/cli/create.py
```diff
@@ -55,12 +57,16 @@ def create(opts):
...
+        init_db=not opts['--no-init-db'],
         )
 
 
 def create_environment(environment_dir, port, ckan_version, create_skin,
         site_name, start_web, create_sysadmin, address, log_syslog=False,
-        datapusher=True, quiet=False, site_url=None):
+        datapusher=True, quiet=False, site_url=None, init_db=True):
+    if not init_db:
+        print 'Since the database will not be initialized, we will not copy datapusher.'
```
wardi: not necessary if explained in comment as above

PR #328, opened 2015-09-17
Fully clone CKAN for master image & Update Documentation

JackMc:


JackMc:
This is a WIP while I write docs

JackMc:
No longer WIP

wardi:
hmm. why is circleci unhappy?

wardi:
I can't re-run the test because the branch is on your repo. Let's start pushing branches to datacats/datacats instead so we can all work on them and restart CI when necessary.

Merging this because tests *should* be fine now

JackMc:
Sounds like a good idea!

PR #337, opened 2015-10-02
move individual tests back in to circle.yml

wardi:


JackMc:
I enjoy test.sh cause I can run it locally

wardi:
test.sh hides the command execution time and prevents circleci from hilighting and emailing the specific commands that failed. I hear what you're saying though. how about a little python script that executes the commands for you like test.sh did?

JackMc:
I think that'd be a fair compromise.

wardi:
Let's leave it as it is, this is an ugly-looking change

PR #342, opened 2015-10-08
Try and fix the service-provider issue

JackMc:
This is what's used to generate the images on Docker Hub.


deniszgonjanin:
My PR to ckan-service-provider just got merged an hour ago. I'm fine with leaving this PR as it is though, @wardi can decide

docker/ckan/Dockerfile
```diff
@@ -10,6 +10,7 @@ RUN virtualenv $CKAN_HOME && \
     chown -R www-data:www-data /var/www && \
     git clone --depth 1 --branch $BRANCH https://github.com/ckan/ckan /project/ckan && \
     git clone --branch stable --depth 1 https://github.com/ckan/datapusher /project/datapusher && \
+    $CKAN_HOME/bin/pip install -e git+https://github.com/deniszgonjanin/ckan-service-provider#egg=ckanserviceprovider && \
```
wardi: anything we install here gets blown away when a user runs `install --clean` so we would need to add this hack there too :-(

deniszgonjanin:
datapusher master's ckan-service-provider requirement got bumped to 0.0.5, which has the SQLAlchemy fix, and a new version of provider was released on pypi. Parts of this PR could be reverted to point to the official repos again.

JackMc:
Yay! Unfortunately we run on data pusher's stable branch :-(

deniszgonjanin:
Ah. Damn

docker/ckan/Dockerfile
wardi: @JackMc this is the last thing holding up this PR, would you confirm that `install --clean` leaves us in a broken state? And if so, other than forcing install of this branch like we do for ckanapi is there another way we could fix this problem?

Issue #334, opened 2015-09-30
Vanilla datacats environment fails to start with AttributeError


florianm:
Ok, here's a tricky one. My system should be up to date with:
* apt-get upgraded Ubuntu 14.04 VM
* fresh datacats master install (1.1.0)
* fresh docker install (updated) via `wget -qO- https://get.docker.com/ | sh`
* fresh datacats images via `datacats pull -a`

I've got a handful of existing datacats environments running on that VM which all work fine and can be stopped/restarted. However, creating a new datacats environment, or running `datacats install` inside an existing one will result in a crashing ckan instance.

Live example: 
`datacats create --ckan latest --site-url http://minesweeper.alpha.data.wa.gov.au minesweeper PORT`
creates http://minesweeper.alpha.data.wa.gov.au/ which initially works fine, but as soon as I login (using the admin account created during `datacats create`), it crashes with error trace:

```
Debug at: http://minesweeper.alpha.data.wa.gov.au/_debug/view/1443596557
2015-09-30 07:04:27,533 INFO  [ckan.lib.base]  /images/icons/ckan.ico render time 0.004 seconds
Traceback (most recent call last):
  ...
File "/project/ckan/ckan/model/activity.py", line 298, in dashboard_activity_list
    q = _dashboard_activity_query(user_id, limit + offset)
  File "/project/ckan/ckan/model/activity.py", line 283, in _dashboard_activity_query
    q1 = _user_activity_query(user_id, limit)
  File "/project/ckan/ckan/model/activity.py", line 126, in _user_activity_query
    return _activities_union_all(q1, q2)
  File "/project/ckan/ckan/model/activity.py", line 96, in _activities_union_all
    return model.Session.query(model.Activity).select_entity_from(
AttributeError: 'Query' object has no attribute 'select_entity_from'
```
Hitting /user/logout restores a working CKAN again.

This happens with both `--ckan latest` and `--ckan 2.4`.

If I run `datacats install` in an existing (previously working) environment, datapusher will install https://github.com/okfn/ckan-service-provider/archive/stable.zip which in turn pulls in SQLAlchemy==0.7.8 which in turn collides with latest (vanilla) ckanext-spatial's philosophy on GeoAlchemy2. Strangely, the relevant test in ckanext-spatial which tests for GeoAlchemy versions has been around since well before I ran my last (working) ckan installs.

[...]

PR #370, opened 2015-12-06
Default to ckan:2.4 image

mchelen:
ckan:2.3 has a critical error that is resolved in 2.4 https://github.com/datacats/datacats/issues/334

### DIFF
--- a/datacats/cli/pull.py
+++ b/datacats/cli/pull.py
@@ -19,7 +19,8 @@ IMAGES = [
 
 EXTRA_IMAGES = [
     'datacats/lessc',
-    'datacats/ckan:latest'
+    'datacats/ckan:latest',
+    'datacats/ckan:2.3'
     ]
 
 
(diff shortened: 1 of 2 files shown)


datacats/cli/pull.py
```diff
@@ -12,15 +12,14 @@
...
+    'datacats/ckan:2.4',
     'datacats/postgres',
     'datacats/solr'
     ]
 
 EXTRA_IMAGES = [
     'datacats/lessc',
-    'datacats/ckan:latest',
-    'datacats/ckan:2.4'
+    'datacats/ckan:latest'
```
wardi: shouldn't this have a `datacats/ckan:2.3' added?

mchelen: I wasn't sure if there was any reason to have the 2.3 version, but I added it back in case anyone needs it.

wardi:
https://github.com/datacats/datacats/blob/master/datacats/cli/create.py#L33 will need to be changed too or this won't change the default

mchelen:
@wardi Ah ok, added to the PR. Does that line change the default value or is it just part of the documentation?

wardi:
@mchelen it does both! docopt is magic (in a good way)

@JackMc it looks like this should work, If you're happy with it please go ahead and merge

